PR #9, opened 2021-02-19
use project id as external-name

### LINKED ISSUE #7: [project]: create project without setting external-name annotation
### What happened?
Creating a `project` without the `crossplane.io/external-name` annotation fails because the `name` is blank and the gitlab api refuses this:
```json
{
   "object":{
      "kind":"Project",
      "name":"example-project",
      "uid":"96170c21-04cc-49e4-b533-1edff54b5aed",
      "apiVersion":"projects.gitlab.crossplane.io/v1alpha1",
      "resourceVersion":"5730"
   },
   "reason":"CannotCreateExternalResource",
   "message":"cannot create Gitlab project: POST https://gitlab.com/api/v4/projects: 400 {message: {limit_reached: []}, {name: [can't be blank, can contain only letters, digits, emojis, '_', '.', dash, space. It must start with letter, digit, emoji or '_'.]}, {path: [can't be blank, can contain only letters, digits, '_', '-' and '.'. Cannot start with '-', end in '.git' or end in '.atom', is too short (minimum is 1 character)]}}"
}
```

`provider-gitlab` should use `metadata.name` as the name.

### How can we reproduce it?
create a project without the `crossplane.io/external-name` annotation:
```yaml
kind: Project
metadata:
  name: example-project
  # annotations:
  #   crossplane.io/external-name: example-project
spec:
  forProvider:
    namespaceId: 12345678
    description: "an example project"
  providerConfigRef:
    name: gitlab-provider
  # a reference to a Kubernetes secret to which the controller will write the runnersToken
  writeConnectionSecretToRef:
    name: gitlab-project-example-project
    namespace: crossplane-system
```

provider-gitlab version: 9c0dd48afddab972940e7bbea29b52a5da6d2232

janwillies:

- use project id instead of path/name combination as `external-name`, which makes creating a project referencer easier (e.g. for project hooks)
- drop tests for updating external-name

Note that this is a breaking change but given that we don't have a release yet I think we can get away with it

Fixes #7 

I have:

manually created a `project` without setting an `annotation` and deleting it again.

### DIFF
--- a/apis/projects/v1alpha1/project_types.go
+++ b/apis/projects/v1alpha1/project_types.go
@@ -419,11 +419,11 @@ type ProjectStatus struct {
 
 // +kubebuilder:object:root=true
 
-// A Project is a managed resource that represents an AWS Elastic Kubernetes
-// Service Project.
+// A Project is a managed resource that represents a Gitlab Project
 // +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
 // +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
 // +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:printcolumn:name="PATH WITH NAMESPACE",type="string",JSONPath=".status.atProvider.pathWithNamespace"
 // +kubebuilder:subresource:status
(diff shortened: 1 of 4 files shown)


Knappek:
LGTM @janwillies . One suggestion: because it is not possible anymore to quickly see `pathWithNamespace` I would suggest to add an additional printercolumn [here](https://github.com/crossplane-contrib/provider-gitlab/blob/master/apis/projects/v1alpha1/project_types.go#L426). What do you think?

janwillies:
Thanks! I can see it being helpful with projects in different groups. That being said, we should probably introduce `name` as parameter, so that two projects in different groups can have the same name

PR #15, opened 2021-03-22
move baseURL to top-level ProviderConfig

### LINKED ISSUE #8: document format of the secret
### What happened?
It's not easy to figure out the connection format for the secret.

```
kubectl create secret generic gitlab-credentials -n crossplane-system --from-literal=credentials="token:YOURTOKEN"
```

### How can we reproduce it?
Try to connect `provider-gitlab` to gitlab.com. There is no documentation on the format of the secret.

provider-gitlab version: 9c0dd48afddab972940e7bbea29b52a5da6d2232

janwillies:

- move baseURL to ProviderConfig and document in readme

it's now easier to target a different gitlab instance:

```yaml
apiVersion: gitlab.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: gitlab-provider
spec:
  baseURL: https://gitlab.com/
  credentials: [...]
```

Fixes #8 

I have:

manual

### DIFF
--- a/pkg/clients/gitlab.go
+++ b/pkg/clients/gitlab.go
@@ -18,9 +18,7 @@ package clients
 
 import (
 	"context"
-	"fmt"
 
-	"github.com/go-ini/ini"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	gitlab "github.com/xanzy/go-gitlab"
@@ -34,9 +32,6 @@ import (
 	"github.com/crossplane-contrib/provider-gitlab/apis/projects/v1alpha1"
 	"github.com/crossplane-contrib/provider-gitlab/apis/v1beta1"
 )
-
-// DefaultSection for INI files.
-const DefaultSection = ini.DefaultSection
 
 // Config provides gitlab configurations for the Gitlab client
 type Config struct {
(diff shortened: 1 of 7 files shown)


pkg/clients/gitlab.go
```diff
@@ -93,39 +87,12 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
 		if err := c.Get(ctx, types.NamespacedName{Namespace: csr.Namespace, Name: csr.Name}, s); err != nil {
 			return nil, errors.Wrap(err, "cannot get credentials secret")
 		}
-		return UseProviderSecret(ctx, s.Data[csr.Key], DefaultSection)
+		return &Config{BaseURL: pc.Spec.BaseURL, Token: string(s.Data[csr.Key])}, nil
```
MisterMX, line 90: `BaseURL` is not marked with `kubebuilder:validation:Required`, so it might be empty. It the provider able to deal with that or should it instead default to `https://gitlab.com`?

janwillies: We are indeed defaulting to gitlab.com in case the `baseURL` is empty: https://github.com/crossplane-contrib/provider-gitlab/blob/master/pkg/clients/gitlab.go#L51-L55

janwillies: (or rather the gitlab sdk is)

MisterMX: OK, then this should resolve it

janwillies: Thanks! I just tested it as well and it works as expected 👍

MisterMX:
LGTM

Issue #10, opened 2021-02-26
project without namespaceid fails


janwillies:
### What happened?
creating a `project` without a `namespaceId` works (I can see the `project` in my personal namespace in gitlab) but provider-gitlab produces the following error:
```json
{
   "reconcilerGroup":"projects.gitlab.crossplane.io",
   "reconcilerKind":"Project",
   "controller":"managed/project",
   "name":"example-project",
   "namespace":"",
   "error":"cannot update managed resource status: Project.projects.gitlab.crossplane.io \"example-project\" is invalid: [status.atProvider.owner.confirmedAt: Invalid value: \"null\": status.atProvider.owner.confirmedAt in body must be of type string: \"null\", status.atProvider.owner.currentSignInAt: Invalid value: \"null\": status.atProvider.owner.currentSignInAt in body must be of type string: \"null\", status.atProvider.owner.identities: Invalid value: \"null\": status.atProvider.owner.identities in body must be of type array: \"null\", status.atProvider.owner.lastSignInAt: Invalid value: \"null\": status.atProvider.owner.lastSignInAt in body must be of type string: \"null\", status.atProvider.owner.lastActivityOn: Invalid value: \"null\": status.atProvider.owner.lastActivityOn in body must be of type string: \"null\", status.atProvider.owner.createdAt: Invalid value: \"null\": status.atProvider.owner.createdAt in body must be of type string: \"null\", status.atProvider.owner.customAttributes: Invalid value: \"null\": status.atProvider.owner.customAttributes in body must be of type array: \"null\"]",
[...]

PR #27, opened 2021-05-20
Project in personal namespace fails. Fix

Stanislav2017:
Signed-off-by: Stanislav2017 <stanislavchichek@gmail.com>

I this PR. I changed register first char of the method name to lower case for "lateInitialize" and "isProjectUpToDate" and moved the methods to the controller as it is a part of it and is used only in it.
Removed func "updateExternalName" and start use standarв crossplane capability for set external name. I used this path of code

kubernet tries to set the value and if the types do not match it returns an error. But if omitempty is added it sets default value for each type

meta.SetExternalName(cr, strconv.Itoa(prj.ID))
return managed.ExternalCreation{ExternalNameAssigned: true}, errors.Wrap(err, errKubeUpdateFailed)

Fixes https://github.com/crossplane-contrib/provider-gitlab/issues/10

I have:

manually


janwillies:
Thanks for the PR @Stanislav2017! Looks very good 👍 Could you please update the description and provide a small summary what you did in this PR? e.g. how did you solve the issue (presumably by adding `omitempty`)? I also see that you moved some functions to the controller and simplified `setExternalName`

Stanislav2017:
> Thanks for the PR @Stanislav2017! Looks very good 👍 Could you please update the description and provide a small summary what you did in this PR? e.g. how did you solve the issue (presumably by adding `omitempty`)? I also see that you moved some functions to the controller and simplified `setExternalName`

ok. I added short description

janwillies:
This looks very good! Thanks a lot for fixing the issue

Issue #23, opened 2021-04-27
add referencers for groups


janwillies:
### What problem are you facing?
When creating a `project` or a `group` which references another group as its `parentId` I want to be able to use kubernetes references instead of hardcoded IDs.

### How could Crossplane help solve your problem?
the `group` resource should implement kubernetes references, like for the `project` resource: https://github.com/crossplane-contrib/provider-gitlab/blob/master/apis/projects/v1alpha1/referencers.go. There should be new types for `group`:
```yaml
parentId: 1234
parentIdRef:
  name: parent-group
parentIdSelector:
  labelName: labelValue
```
similar for `projects`:
```yaml
namespaceId: 1234
namespaceIdRef:
  name: parent-group
namespaceIdSelector:
  labelName: labelValue
```

PR #31, opened 2021-07-02
Add referencers for group v2

Stanislav2017:
Signed-off-by: Stanislav2017 stanislavchichek@gmail.com

Added referencers for relation SubGroup with Group and Project with Group without hardcoded IDs

Fixes https://github.com/crossplane-contrib/provider-gitlab/issues/23

I have:

manually

### DIFF
--- a/examples/groups/group.yaml
+++ b/examples/groups/group.yaml
@@ -4,9 +4,10 @@ metadata:
   name: example-group
 spec:
   forProvider:
-    parentId: <parent-group-id>
-    path: "example-path"
-    description: "example description"
+    parentIdRef:
+      name: example-parent-group
+    path: "example-group-path"
+    description: "example group description"
   providerConfigRef:
     name: gitlab-provider
   # a reference to a Kubernetes secret to which the controller will write the runnersToken
--- a/apis/groups/v1alpha1/group_types.go
+++ b/apis/groups/v1alpha1/group_types.go
@@ -17,8 +17,9 @@ limitations under the License.
 package v1alpha1
 
 import (
-	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 )
 
 // VisibilityValue represents a visibility level within GitLab.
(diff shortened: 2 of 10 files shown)


examples/groups/group.yaml
```diff
@@ -2,11 +2,12 @@ apiVersion: groups.gitlab.crossplane.io/v1alpha1
 kind: Group
 metadata:
   name: example-group
+  annotations:
+    crossplane.io/external-name: "GROUP ID"
```
janwillies, line 6: Is there a specific reason the `annotation` was added to the example? If it's not strictly needed, I believe we should omit it

examples/groups/subgroup.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: groups.gitlab.crossplane.io/v1alpha1
+kind: Group
+metadata:
+  name: example-subgroup
```
janwillies, line 4: Since this is just a `group`, should we really add it as another example? My preference is to omit this example

examples/projects/groupproject.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: projects.gitlab.crossplane.io/v1alpha1
+kind: Project
+metadata:
+  name: example-group-project
```
janwillies, line 4: same as with the subgroup example, since this is just a `project` my preference is to omit this example. People can take a look at the API documentation at gitlab or doc.crds.dev/crossplane-contrib/provider-gitlab (once it is released) to find all features.

janwillies (changes requested):
thanks @Stanislav2017! Small nits about the `examples/` but can be merged afterwards

examples/groups/group.yaml
Stanislav2017: I think the annotation is necessary, since there is no way to create top-level groups, we can only load a group from gitlab by id

examples/groups/subgroup.yaml
Stanislav2017: I can, but need to add this commented text 
parentIdRef:
      name: example-group
OR
namespaceIdRef:
      name: example-project
to group or project

examples/groups/group.yaml
Stanislav2017: Removed annotations

examples/groups/subgroup.yaml
Stanislav2017: Removed examples/groups/subgroup.yaml

examples/projects/groupproject.yaml
Stanislav2017: Removed examples/projects/groupproject.yaml

examples/groups/group.yaml
```diff
@@ -4,9 +4,10 @@ metadata:
   name: example-group
 spec:
   forProvider:
-    parentId: <parent-group-id>
-    path: "example-path"
-    description: "example description"
+    parentIdRef:
+      name: example-group
```
janwillies, line 8: suggestion: 
`name: example-parent-group`
to make it clear it's a different group we are referencing.

janwillies:
Thanks a lot @Stanislav2017!

PR #42, opened 2021-07-30
Add name field to group and project

### LINKED ISSUE #41: Group and Project Name fields
Currently, `Project` and `Group` take the Kubernetes resource name to be the name of the project or group in Gitlab. 

This works fine, but the Kubernetes resource name format is much more restrictive than the actual name field in Gitlab (which is effectively a display name for human consumption - and can include spaces and uppercase).

Right now, we're restricting managed groups and projects to having display names containing all lowercase characters and no spaces.

Add a `name` or `displayName` field as part of the spec of any relevant resources which, when set, is used instead of the Kubernetes resource name when managing the resource.

Happy to put in a PR for this if this is deemed worthwhile 👍

benagricola:

Adds a new `name` field to `Group` and `Project` resources - this allows users to set a 'human readable' name for their `Group` or `Project` that is free of the restrictions placed on the `metadata.name` of a resource.

Closes #41

I have:

An additional test has been added to each resource to validate the override behaviour of the `name` field.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ crds.clean:
 	@find package/crds -name *.yaml.sed -delete || $(FAIL)
 	@$(OK) cleaned generated CRDs
 
-generate: crds.clean
+generate.done: crds.clean
 
 # Ensure a PR is ready for review.
 reviewable: generate lint
--- a/apis/groups/v1alpha1/group_types.go
+++ b/apis/groups/v1alpha1/group_types.go
@@ -73,6 +73,12 @@ type GroupParameters struct {
 	// +optional
 	Description *string `json:"description,omitempty"`
 
+	// Name is the human-readable name of the group.
+	// If set, it overrides metadata.name.
+	// +kubebuilder:validation:MaxLength:=255
+	// +optional
+	Name *string `json:"name,omitempty"`
+
 	// Prevent adding new members to project membership within this group.
 	// +optional
 	MembershipLock *bool `json:"membershipLock,omitempty"`
--- a/apis/groups/v1alpha1/zz_generated.deepcopy.go
+++ b/apis/groups/v1alpha1/zz_generated.deepcopy.go
@@ -300,6 +300,11 @@ func (in *GroupParameters) DeepCopyInto(out *GroupParameters) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.Name != nil {
+		in, out := &in.Name, &out.Name
+		*out = new(string)
+		**out = **in
+	}
 	if in.MembershipLock != nil {
 		in, out := &in.MembershipLock, &out.MembershipLock
 		*out = new(bool)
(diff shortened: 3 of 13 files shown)


janwillies:
Hey @benagricola, thanks for the PR! 

Looking at the [Gitlab API](https://docs.gitlab.com/ee/api/projects.html#create-project) it seems we should name the field `name`. Wonder what your thoughts are for naming it `displayName`, I see in the associated ticket you considered both options.

benagricola:
> Hey @benagricola, thanks for the PR!
> 
> Looking at the [Gitlab API](https://docs.gitlab.com/ee/api/projects.html#create-project) it seems we should name the field `name`. Wonder what your thoughts are for naming it `displayName`, I see in the associated ticket you considered both options.

Morning, @janwillies - I don't really have a good answer for you here 😄 I thought it was more clear being `displayName` rather than `name`, as I tend to associate `name` with the `metadata.name` of resources which is more like an identifier than a "human name".

It is of course not "high fidelity" in the Crossplane sense so I'm happy to search and replace the name of the field if you're generally happy with the approach taken other than that.

janwillies:
My thinking was that it might be confusing for new users searching the gitlab api doc and not finding `displayName`. So yeah, I have a slight preference for `name` to match the gitlab api as close as possible (if there are no other disadvantages)

Approach is definitely good! Would be happy to accept the PR

benagricola:
> My thinking was that it might be confusing for new users searching the gitlab api doc and not finding `displayName`. So yeah, I have a slight preference for `name` to match the gitlab api as close as possible (if there are no other disadvantages)
> 
> Approach is definitely good! Would be happy to accept the PR

Excellent 👍 - S&R force push coming soon

benagricola:
@janwillies should be good to go now!

janwillies:
lgtm!

I wonder why the e2e tests fail though 🤔

benagricola:
> lgtm!
> 
> I wonder why the e2e tests fail though 🤔

I think this might be an issue with the e2e tests rather than this specific PR.

It looks similar to a recent issue on `provider-aws`: https://github.com/crossplane/provider-aws/pull/782#issuecomment-888612043 - I'm going to try running locally with kind `v0.11.1` to see if it works.

benagricola:
@janwillies I think bumping the build submodule version in the most recent commit will fix the e2e issue - it does locally at least.

hasheddan (commented):
@benagricola you can fix issues here by changing the following line:
https://github.com/crossplane-contrib/provider-gitlab/blob/6c75a45e6dbf3301f680e12354b4695776707ac8/Makefile#L76
to 
```
generate.done: crds.clean
```

hasheddan (commented):
@benagricola ah also this change: https://github.com/crossplane/provider-aws/pull/645/files

janwillies:
Thanks a lot for pushing this through @benagricola !

PR #45, opened 2021-08-05
Make sure project and group name is checked for updates

benagricola:
Forgot to implement up to date checking of the new `name` fields added in #42.

This adds additional checks to the `upToDate` methods, and adds a test for it on the `Project` resource.

I can't for the life of me work out how to add a suitable test within the existing framework for the `Group` resource - regardless of what I set, the test always passes when it should fail without the fixes to the `upToDate` method. Any ideas @janwillies ?

I have:

Additional test added for Project. Group change is untested but identical.

### DIFF
--- a/pkg/controller/groups/group.go
+++ b/pkg/controller/groups/group.go
@@ -164,6 +164,9 @@ func (e *external) Delete(ctx context.Context, mg resource.Managed) error {
 
 // isGroupUpToDate checks whether there is a change in any of the modifiable fields.
 func isGroupUpToDate(p *v1alpha1.GroupParameters, g *gitlab.Group) bool { // nolint:gocyclo
+	if p.Name != nil && !cmp.Equal(*p.Name, g.Name) {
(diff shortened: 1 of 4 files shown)


benagricola:
Issues with the `Group` test look to be a problem with the `Visibility` `upToDate` check as it has a value set, but is checking against the empty value of `gitlabGroup`.

I need to have a think around how to fix this.

benagricola:
Latest commit should fix all of these issues. Essentially, missing values for the mock returned `gitlab.Group` would cause `upToDate` to return `false` regardless of what fields were being tested (so the existing test results were not accurate).

This also adds missing `upToDate` checks for `projectCreationLevel` and `subGroupCreationLevel`.

I've tested each of the branches of the method (new and old) to confirm that removing each branch causes the relevant test to fail.

I think this is good to go.

janwillies:
LGTM

Thanks a lot @benagricola!

PR #47, opened 2021-08-10
Add Project CI Variables

### LINKED ISSUE #46: Implement Project / Group Variables
We need to be able to set project (and potentially group) level CI/CD variables to control deployments of template-imported projects.

Implement a new resource type, `Variable` or similar which uses the [Variables API](https://docs.gitlab.com/ee/api/project_level_variables.html) to CRUD variables on a project or group.

A Hi-fi implementation would suggest to use two types, e.g. `Variable` under each API group, which resolves the appropriate project or group, and potentially resolve a secret, or simply take a plaintext value to set the variable.

I was thinking of an initial API like the following (replace `project` with `group` for the second type):

```yaml
kind: Variable
metadata:
  name: deploy-arn
spec:
  forProvider:
    projectIdRef:
      name: my-project
    # projectId: 1234
    # projectIdSelector:
    #   matchLabels:
    #     project-type: blah
    type: env_var # (default) or file
    protected: true
    masked: false
    environmentScope: "dev/*"
    key: AWS_ROLE_ARN
    value: arn:aws:iam::999999999:role/my-deploy-role
```

Happy to implement this if the API design looks good.

benagricola:
This PR implements GitLab [Project Variables](https://docs.gitlab.com/ee/api/project_level_variables.html), allowing Crossplane to manage project-level CI/CD variables and exert control over the build process.

Currently this only supports plain-text variable values so should not be used with sensitive content, although it does support protecting and masking values once in GitLab.

I've taken a slightly different approach to implementation of `upToDate` in this resource, converting the `gitlab.ProjectVariable` representation back into our own `v1alpha1.VariableParameters` struct, and then comparing the two with `cmp.Equal`. I've found this approach to be robust and more succinct in `provider-cloudflare`, but I know this doesn't follow convention within this provider so happy to implement differently if necessary.

Adding the same functionality for Group Variables should be a simple copy-and-paste - happy to add to this PR, or submit separately. I've not done this yet as I think the new code will be easier to review without it.

I have:

Tests have been added for the additional controller and client, and the provider itself has been run against a `kind` cluster and a manually created project to confirm functionality against real GitLab.


apis/projects/v1alpha1/variable_types.go
```diff
@@ -0,0 +1,121 @@
...
+	Masked *bool `json:"masked,omitempty"`
+
+	// Protected enables or disables variable protection.
+	// +optional
+	Protected *bool `json:"protected,omitempty"`
+
+	// VariableType is the type of the variable.
+	// +kubebuilder:validation:Enum:=env_var;file
+	// +optional
+	VariableType *VariableType `json:"variableType,omitempty"`
```
negz, line 73: Is it worth defaulting this to `env_var`?

apis/projects/v1alpha1/variable_types.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// EnvironmentScope indicates the environment scope
+	// that this variable is applied to.
+	// +optional
+	EnvironmentScope *string `json:"environmentScope,omitempty"`
+}
+
+// VariableObservation represents a CI variable.
+type VariableObservation struct{}
```
negz, line 82: We can probably just omit this if it's unused. (Unless that makes angryjet think this type isn't a managed resource? I forget.)

pkg/controller/projects/variables/controller.go
```diff
@@ -0,0 +1,150 @@
...
+func SetupVariable(mgr ctrl.Manager, l logging.Logger) error {
+	name := managed.ControllerName(v1alpha1.VariableKind)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&v1alpha1.Variable{}).
+		Complete(managed.NewReconciler(mgr,
+			resource.ManagedKind(v1alpha1.VariableGroupVersionKind),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newGitlabClientFn: projects.NewVariableClient}),
+			managed.WithInitializers(managed.NewDefaultProviderConfig(mgr.GetClient())),
```
negz, line 58: Nit: If you wanted, you could bump to use the latest `crossplane-runtime` and not pass in the `NewDefaultProviderConfig`. Building against the latest runtime will generate CRDs that use OpenAPI defaulting of `spec.providerConfigRef.name`, which saves us an `Update` call.

pkg/controller/projects/variables/controller.go
```diff
@@ -0,0 +1,150 @@
...
+	kube   client.Client
+	client projects.VariableClient
+}
+
+func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
+	cr, ok := mg.(*v1alpha1.Variable)
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errNotVariable)
+	}
+	variable, _, err := e.client.GetVariable(*cr.Spec.ForProvider.ProjectID, cr.Spec.ForProvider.Key)
```
negz, line 90: Hrm, I guess external-name is kind of unintuitive in this context? I imagine it would technically be the variable key?

negz (approved):
Looks good to me - none of my comments are blocking. I'd prefer to have someone like @janwillies who knows this provider better approve before we merge though.

apis/projects/v1alpha1/variable_types.go
benagricola: I didn't do this as `variable_type` is an optional field when passed to the GitLab API, and is defaulted on that side and then late initialised if not set. 

Happy to default this on our side, but I was under the impression we generally wanted to rely on default behaviour of the remote API as much as possible.

pkg/controller/projects/variables/controller.go
benagricola: @negz If not passing in the `NewDefaultProviderConfig`, ISTM that the `WithInitializers()` call is redundant too?

pkg/controller/projects/variables/controller.go
benagricola: Yeah this was my thinking. There isn't a single unique value to identify a variable (it is Project ID + Key), so putting the Key into the external-name field is just adding another level potential breakage.

janwillies:
lgtm, thanks @benagricola!

would you please squash the commits?

benagricola:
Done @janwillies :)

PR #69, opened 2023-03-22
feat(variable): optionally get value from secret

### LINKED ISSUE #64: update xanzy/go-gitlab to latest release
We are currently using version [v0.50.0](https://github.com/crossplane-contrib/provider-gitlab/blob/master/go.mod#L10) of the xanzy/go-gitlab sdk, which is quite dated by now

Update to the [latest release](https://github.com/xanzy/go-gitlab/tags) and add new features

### LINKED ISSUE #67: can't pass sensitive data to Variable in an secure manner

We would like to use crossplane to create on our k8s namespace SA Token and set it as a [Project Variable](https://doc.crds.dev/github.com/crossplane-contrib/provider-gitlab/projects.gitlab.crossplane.io/Variable/v1alpha1@v0.2.0)

Both Variable and SA will be created in same Composition but if we would like to pass value between Composed Resources we would need to use [ToCompositeFieldPath](https://docs.crossplane.io/v1.11/concepts/composition/#patching-from-one-composed-resource-to-another) which means setting copying k8s Secret on value to CompositeResource annotation (which does not seem to be safe)

It would be nice to have on [Variable](https://doc.crds.dev/github.com/crossplane-contrib/provider-gitlab/projects.gitlab.crossplane.io/Variable/v1alpha1@v0.2.0) field `spec.forProvider.valueFromSecret`

somebadcode:
* Updated `xanzy/go-gitlab` to v0.81.0 to get support for the `raw` variable field.
* Fixed deprecation warnings introduced by newer `xanzy/go-gitlab`.
* New field `v1alpha1.Variable.ValueSecretRef` allows you to get the value from a Kubernetes secret.
* `v1alpha1.Variable.Value` is now optional and is instead mutually exclusive to the new field `ValueSecretRef`.

Added the option to get a variable's value from a Kubernetes secret. To achieve this I had to upgrade the version of  `xanzy/go-gitlab`used since the use of the `raw` option in the API becomes more important when working with masked values, which is automatically done when getting the value from a secret. Also made the necessary changes required by this upgrade.

Fixes #64
Fixes #67 

I have:

Tested on a local cluster during development. Added tests for the new use cases.
Unable to complete the e2e tests. Everything looks good but the test never finishes since it's perpetually waiting for the     provider to be deleted despite it not being visible in the Kind cluster.


MisterMX (approved):
LGTM. Thanks a lot @somebadcode.

PR #72, opened 2023-04-05
fix(e2e): change while loop that waits for deletion

### LINKED ISSUE #71: end-to-end tests do not work on MacOS
### What happened?

When running `make e2e`, everything goes well until it reaches the point where it waits for all resources of type `providerrevision.pkg.crossplane.io` to get deleted. The comparison in the loop is not valid on MacOS X.

### How can we reproduce it?

Run `make e2e` on MacOS X using stock shell and `wc`.

Crossplane version: 1.11.x
Crossplane Provider GitLab version: 0.3.0
OS: MacOS X 13.3
Kubernetes: 1.22.17

somebadcode:
Changed the condition for the while loop that waits for the number of `providerrevision.pkg.crossplane.io` to reach 0 to complete the test.

Changed the expression in the condition of the while loop that waits for the deletion of the provider to complete.
Instead of waiting for the result to be a literal string with the value `"0"`, it will first convert it to an integer. The conversion to an integer will strip away any white spaces. On MacOS X, the command returns the correct value but it has leading white spaces which causes the loop to become infinite. Converting it to an integer first gets around this issue.

Fixes #71 

I have:

I've run `make e2e`.

### DIFF
--- a/cluster/local/integration_tests.sh
+++ b/cluster/local/integration_tests.sh
@@ -187,17 +187,10 @@ echo_step "uninstalling ${PROJECT_NAME}"
 
 echo "${INSTALL_YAML}" | "${KUBECTL}" delete -f -
 
-# check pods deleted
-timeout=60
-current=0
-step=3
-while [[ $(kubectl get providerrevision.pkg.crossplane.io -o name | wc -l) != "0" ]]; do
-  echo "waiting for provider to be deleted for another $step seconds"
-  current=$current+$step
-  if ! [[ $timeout > $current ]]; then
-    echo_error "timeout of ${timeout}s has been reached"
-  fi
-  sleep $step;
-done
+# wait for pods to be deleted
+echo "waiting for provider to be deleted (times out after 60 seconds)"
+if ! kubectl wait providerrevision.pkg.crossplane.io --for=delete --all --timeout=60s ; then
+  echo_error "Waiting for deletion of providers failed"
+fi
 
 echo_success "Integration tests succeeded!"


cluster/local/integration_tests.sh
```diff
@@ -191,7 +191,7 @@ echo "${INSTALL_YAML}" | "${KUBECTL}" delete -f -
 timeout=60
 current=0
 step=3
-while [[ $(kubectl get providerrevision.pkg.crossplane.io -o name | wc -l) != "0" ]]; do
+while [[ ! $(($(kubectl get providerrevision.pkg.crossplane.io -o name | wc -l))) ]]; do
```
MisterMX, line 194: If you just want to check the exit code, I believe you can just do 

```bash
while ! kubectl get providerrevision.pkg.crossplane.io -o name | wc -l; do
```

However, that doesn't work here since the exit code is always zero. You would have to do something like

```bash
while [ "$(kubectl get providerrevision.pkg.crossplane.io -o name | wc -l)" != "0" ]; do
```

That should also work with `/bin/sh`.

MisterMX (changes requested):

cluster/local/integration_tests.sh
somebadcode: The problem is that the string `"0"` does not work on MacOS X. `wc` prints the count with whitespace. For example `"        0"`. So a simple string compare doesn't work unless you trim the spaces one way or another. The loop keeps going until there are 0 providers running.

It captures stdout and checks if it's 0. The original code compares a string (broken when using MacOS X `wc`) but converting it to an integer will remove the spaces and the truthiness of the resulting integer can be used instead of the string.

somebadcode: `$(( "$x" ))` converts `x` to an integer.

It's 100% valid Bash syntax. The script's header says that it's a Bash script and it already used double brackets. The trimming of whitespaces is the goal of my change. The solution works on all systems that have Bash installed.

MisterMX: Then I would suggest a K8s native way like
```bash
kubectl wait providerrevision.pkg.crossplane.io --for=delete --all
```

somebadcode: Good point. I changed it to use that instead and it works fine.

cluster/local/integration_tests.sh
```diff
@@ -187,17 +187,11 @@ echo_step "uninstalling ${PROJECT_NAME}"
...
-    echo_error "timeout of ${timeout}s has been reached"
-  fi
-  sleep $step;
-done
-
-echo_success "Integration tests succeeded!"
+# wait for pods to be deleted
+echo "waiting for provider to be deleted (times out after 60 seconds)"
+kubectl wait providerrevision.pkg.crossplane.io --for=delete --all --timeout=60s
+if [ $? -ne 0 ]; then
```
MisterMX, line 193: Non-zero exit codes are better checked with `if !` rather than `$?`.

```suggestion
if ! kubectl wait providerrevision.pkg.crossplane.io --for=delete --all --timeout=60s ; then
```

MisterMX (approved):
@somebadcode I took the freedom to apply my suggestions. Thank you for your contribution!

PR #75, opened 2023-06-16
Project Access Token Controller 

IntaSlapna:

Fixes #

I have:

- [ ] Read and followed Crossplane's [contribution process].


pkg/controller/projects/accesstokens/controller.go
```diff
@@ -0,0 +1,237 @@
...
+
+func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
+	cr, ok := mg.(*v1alpha1.AccessToken)
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errNotAccessToken)
+	}
+
+	externalName := meta.GetExternalName(cr)
+	if externalName == "" {
+		return managed.ExternalObservation{ResourceExists: false}, nil
```
MisterMX, line 96: `false` is the default value.

```suggestion
		return managed.ExternalObservation{}, nil
```

pkg/controller/projects/accesstokens/controller.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	projectAccessTokenID, err := strconv.Atoi(externalName)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.New(errNotAccessToken)
+	}
+
+	listProjectAccessTokensOptions := gitlab.ListProjectAccessTokensOptions{}
+	accesstokenArr, _, err := e.client.ListProjectAccessTokens(*cr.Spec.ForProvider.ProjectID, &listProjectAccessTokensOptions)
```
MisterMX, line 105: possible nil deref here: `cr.Spec.ForProvider.ProjectID` is an optional field. You'd have to check if it defined first and error out if that's the case.

(Maybe change the description of the field in `apis` to tell the user that the field or its *ref Fields are required.

pkg/controller/projects/accesstokens/controller.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.ExternalCreation, error) {
+	cr, ok := mg.(*v1alpha1.AccessToken)
+	if !ok {
+		return managed.ExternalCreation{}, errors.New(errNotAccessToken)
+	}
+
+	dt, _, err := e.client.CreateProjectAccessToken(
+		*cr.Spec.ForProvider.ProjectID,
```
MisterMX, line 136: possible nil deref here: `cr.Spec.ForProvider.ProjectID` is an optional field. You'd have to check if it defined first and error out if that's the case.

pkg/controller/projects/accesstokens/controller.go
```diff
@@ -0,0 +1,237 @@
...
+
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errCreateFailed)
+	}
+
+	connectionDetails := managed.ConnectionDetails{}
+	connectionDetails["token"] = []byte(dt.Token)
+
+	meta.SetExternalName(cr, strconv.Itoa(dt.ID))
+	return managed.ExternalCreation{ExternalNameAssigned: true, ConnectionDetails: connectionDetails}, nil
```
MisterMX, line 149: Make things more simple:

```suggestion
	meta.SetExternalName(cr, strconv.Itoa(dt.ID))
	return managed.ExternalCreation{
	  ExternalNameAssigned: true,
	  ConnectionDetails: managed.ConnectionDetails{
	    "token": []byte(dt.Token),
	  },
	}, nil
```

MisterMX (changes requested):
@IntaSlapna thanks a lot for your contribution. This looks already pretty good! Please check the remarks I made and fix them. Also, please implement the units test for the controller as well.

cmd/provider/__debug_bin2771274270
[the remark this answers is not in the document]
IntaSlapna: Can't find it locally.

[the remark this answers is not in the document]
MisterMX: It is either still there or you need to commit the deletion of the file.

PR #79, opened 2023-07-11
fix(projects): Avoid setting MirrorUserID in lateInitialize() if atProvider value is 0 (default value)

### LINKED ISSUE #74: Updating a Project causes GitLab API failure (HTTP 400: `mirror_user_id` attribute `is invalid`)
### What happened?

When authenticated as a non-GitLab-administrator user, given a managed resource of kind `Project`, when attempting to modify the managed resource, then the attempt will fail with a HTTP 400 response from the GitLab API relating to the `mirror_user_id` attribute (supposedly an "administrator-only" attribute).

Although this scenario may be an indirect consequence of the GitLab Projects API's behaviour around the `mirror_user_id` attribute, this Crossplane provider could at least gracefully handle the scenario, e.g. potentially skip the attribute somehow.

### How can we reproduce it?

Step 1: create a new `Provider` managed resource.

```sh
$ kubectl apply -f - <<EOF
---
kind: Project
metadata:
  name: created-via-crossplane
spec:
  forProvider:
    namespaceId: <some-existing-group-id>
  providerConfigRef:
    name: <your-gitlab-provider-with-non-admin-user>
EOF
project.projects.gitlab.crossplane.io/created-via-crossplane created

$ kubectl get events | grep project/created-via-crossplane
40s         Normal    CreatedExternalResource        project/created-via-crossplane                  Successfully requested creation of external resource
```

Step 2: update the `Provider` managed resource (via any attribute, e.g. `description`).

```sh
$ kubectl patch project.projects.gitlab.crossplane.io/created-via-crossplane --type=json -p '[{"op": "replace", "path": "/spec/forProvider/description", "value": "crossplane-edited"}]'
project.projects.gitlab.crossplane.io/created-via-crossplane patched

$ kubectl get events | grep project/created-via-crossplane
11m         Normal    CreatedExternalResource        project/created-via-crossplane                  Successfully requested creation of external resource
16s         Warning   CannotUpdateExternalResource   project/created-via-crossplane                  cannot update Gitlab project: PUT https://gitlab.com/api/v4/projects/<project-id>: 400 {message: {mirror_user_id: [is invalid]}}
```

Crossplane version: `v1.12.1`
Crossplane Provider GitLab version: `v0.3.0`

### Related notes

The [GitLab Projects API's `PUT /projects/:id` operation](https://docs.gitlab.com/ee/api/projects.html#edit-project) documents that the `mirror_user_id` attribute is a **non-required** attribute that is also `(administrators only)`. This might explain why the API behaves the way it does, when authenticated via a non-administrator user/token.

Notice that directly `curl`ing the GitLab API yields the same error message:

```sh
$ curl --header "Authorization: Bearer $YOUR_GITLAB_TOKEN" --request PUT https://gitlab.com/api/v4/projects/<project-id> --data "description=laptop-edited" --silent --output /dev/null --write-out "%{response_code}"
200

$ curl --header "Authorization: Bearer $YOUR_GITLAB_TOKEN" --request PUT https://gitlab.com/api/v4/projects/<project-id> --data "description=laptop-edited&mirror_user_id=0" --silent
{"message":{"mirror_user_id":["is invalid"]}}

$ curl --header "Authorization: Bearer $YOUR_GITLAB_TOKEN" --request PUT https://gitlab.com/api/v4/projects/<project-id> --data "description=laptop-edited&mirror_user_id=123" --silent
{"message":{"mirror_user_id":["is invalid"]}}

$ curl --header "Authorization: Bearer $YOUR_GITLAB_TOKEN" --request PUT https://gitlab.com/api/v4/projects/<project-id> --data "description=mirror_user_id=0" --silent
{"error":"allow_merge_on_skipped_pipeline, analytics_access_level, [...], requirements_access_level are missing, at least one parameter must be provided"}
```

Note also: while the [GitLab Groups API's `PUT /groups/:id` operation](https://docs.gitlab.com/ee/api/groups.html#update-group) similarly has a `shared_runners_minutes_limit` attribute that is also non-required and administrator-only, the API silently ignores the attribute, which ultimately means it does not cause errors - whether via `curl` or whether in the context of this Crossplane provider.

```sh
$ curl --header "Authorization: Bearer $YOUR_GITLAB_TOKEN" --request PUT https://gitlab.com/api/v4/groups/<some-group-id> --data "description=laptop-edited&shared_runners_minutes_limit=123" --silent --output /dev/null --write-out "%{response_code}"
200

$ curl --header "Authorization: Bearer $YOUR_GITLAB_TOKEN" --request GET https://gitlab.com/api/v4/groups/<some-group-id> --silent | jq . | grep "shared_runners_minutes_limit"
  "shared_runners_minutes_limit": null,
  "extra_shared_runners_minutes_limit": null,
```

lvvorovi:
Fixes #74 

For already created resources it will require to remove "MirrorUserID" value from managed resource.

Unit Test

### DIFF
--- a/pkg/controller/projects/project.go
+++ b/pkg/controller/projects/project.go
@@ -216,7 +216,7 @@ func lateInitialize(in *v1alpha1.ProjectParameters, project *gitlab.Project) { /
 	if in.MirrorTriggerBuilds == nil {
 		in.MirrorTriggerBuilds = &project.MirrorTriggerBuilds
 	}
-	if in.MirrorUserID == nil {
+	if in.MirrorUserID == nil && project.MirrorUserID != 0 { // since project.MirrorUserID is non-nullable, value `0` treated as `not set`
 		in.MirrorUserID = &project.MirrorUserID
 	}
 	if in.OnlyAllowMergeIfAllDiscussionsAreResolved == nil {
--- a/pkg/controller/projects/project_test.go
+++ b/pkg/controller/projects/project_test.go
@@ -118,6 +118,10 @@ func withAnnotations(a map[string]string) projectModifier {
 	return func(p *v1alpha1.Project) { meta.AddAnnotations(p, a) }
 }
 
+func withMirrorUserIDNil() projectModifier {
+	return func(p *v1alpha1.Project) { p.Spec.ForProvider.MirrorUserID = nil }
+}
+
 func project(m ...projectModifier) *v1alpha1.Project {
 	cr := &v1alpha1.Project{}
 	for _, f := range m {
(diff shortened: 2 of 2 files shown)


pkg/clients/projects/project_test.go
```diff
@@ -422,7 +423,7 @@ func TestGenerateObservation(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			got := GenerateObservation(tc.args.p)
 			if diff := cmp.Diff(tc.want, got); diff != "" {
-				t.Errorf("r: -want, +got:\n%s", diff)
```
MisterMX, line 425: Please keep this as it is.

MisterMX:
@theipster can you check if that fixes your issue?

theipster:
@lvvorovi @MisterMX Thanks for picking this up, and I'm more than happy to check. 🙂 

Silly question, but what's the best way to deploy this PR build into my cluster? I don't see any published artifacts (OCI images? In Docker Hub vs GHCR?) that I can reference...

MisterMX:
> Silly question, but what's the best way to deploy this PR build into my cluster? I don't see any published artifacts (OCI images? In Docker Hub vs GHCR?) that I can reference...

@theipster for quick debugging I would check out the commit of this PR and run the provider locally with the appropriate kube context.

MisterMX (approved):
LGTM. Thank you very much @lvvorovi!

PR #95, opened 2023-08-09
fix(projects.deploykey): Add nil check before pointer deref for CreatedAt in controller

lvvorovi:

- Add nil check before pointer deref for CreatedAt in controller

I have:

### DIFF
--- a/pkg/controller/projects/deploykeys/controller.go
+++ b/pkg/controller/projects/deploykeys/controller.go
@@ -15,7 +15,6 @@ import (
 	"github.com/xanzy/go-gitlab"
 
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
 	controller "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


pkg/controller/projects/deploykeys/controller.go
```diff
@@ -116,9 +116,9 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	lateInitializeProjectDeployKey(&cr.Spec.ForProvider, dk)
 	isLateInitialized := !cmp.Equal(currentState, &cr.Spec.ForProvider)
 
-	cr.Status.AtProvider = v1alpha1.DeployKeyObservation{
-		ID:        &dk.ID,
-		CreatedAt: &metav1.Time{Time: *dk.CreatedAt},
+	cr.Status.AtProvider = v1alpha1.DeployKeyObservation{ID: &dk.ID}
+	if dk.CreatedAt != nil {
+		cr.Status.AtProvider.CreatedAt = &metav1.Time{Time: *dk.CreatedAt}
```
MisterMX, line 121: Instead of an if block you can use a helper function like https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/clients/pointer.go#L123

MisterMX (approved):
LGTM. Thank you very much @lvvorovi!

PR #101, opened 2023-08-10
fix(projects.member): Remove use of 'externalName' to store 'ProjectID' field value

lvvorovi:

- Remove use of 'externalName' to store 'ProjectID' field value

I have:


MisterMX (approved):
LGTM. I think it makes sense to not use the external name here since the membership is identified by the combination of `projectId` and `userId`.

Thank you very much @lvvorovi!

PR #137, opened 2024-04-26
Beta mgmt policy

### LINKED ISSUE #136: Crossplane 1.15 trying to use Management Policies
### What happened?

I'm trying to import a gitlab group so it can be utilized by references

```
kind: Group
metadata:
  name: zys
  annotations:
    crossplane.io/external-name: 'zys'
spec:
  managementPolicies: ['Observe']
  forProvider:
    path: 'xyz/zys'

```

But it get this status:

```
atProvider: {}
conditions:
  - lastTransitionTime: 2024-04-25T12:20:39Z
    message: "`spec.managementPolicies` is set to a non-default value but the
      feature is not enabled: [Observe]"
    reason: ReconcileError
    status: "False"
    type: Synced
```

I believe that is caused by the Gitlab still using the `EnableAlphaManagementPolicies` instead of the `EnableBetaManagementPolicies` now that it's gone into Beta.

### How can we reproduce it?

Try and import an existing group with the Observe managementPolicy set.

Crossplane version:  1.15.2
Crossplane Provider GitLab version: 0.6.1

$ kubectl version
Client Version: v1.30.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.29.3-eks-adc7111

jwhitcraft:

This enabled the using the Beta Management Policy Feature Flag

Fixes #136

I have:

Built and tested in our test controlplane

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -49,7 +49,7 @@ func main() {
 		debug            = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
 		syncInterval     = app.Flag("sync", "Sync interval controls how often all resources will be double checked for drift.").Short('s').Default("1h").Duration()
 		pollInterval     = app.Flag("poll", "Poll interval controls how often an individual resource should be checked for drift.").Default("1m").Duration()
-		leaderElection   = app.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").Bool()
+		leaderElection   = app.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").Envar("LEADER_ELECTION").Bool()
 		maxReconcileRate = app.Flag("max-reconcile-rate", "The global maximum rate per second at which resources may checked for drift from the desired state.").Default("10").Int()
 
 		namespace                  = app.Flag("namespace", "Namespace used to set as default scope in default secret store config.").Default("crossplane-system").Envar("POD_NAMESPACE").String()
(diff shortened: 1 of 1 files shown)


MisterMX:
The provider uses https://github.com/crossplane/crossplane-runtime `v1.14.x` which is using the alpha feature flag. As long as we don't make the upgrade to v1.15 this change wont work.

jwhitcraft:
> The provider uses https://github.com/crossplane/crossplane-runtime `v1.14.x` which is using the alpha feature flag. As long as we don't make the upgrade to v1.15 this change wont work.

From the looks of it, the feature went beta in 1.14.0rc1: https://github.com/crossplane/crossplane-runtime/commit/55a8c109304f3a03509216fb366230d965236c40

cmd/provider/main.go
```diff
@@ -122,8 +122,8 @@ func main() {
 	}
 
 	if *enableManagementPolicies {
-		o.Features.Enable(features.EnableAlphaManagementPolicies)
-		log.Info("Alpha feature enabled", "flag", features.EnableAlphaManagementPolicies)
+		o.Features.Enable(feature.EnableBetaManagementPolicies)
+		log.Info("Beta Management Policies feature enabled", "flag", feature.EnableBetaManagementPolicies)
```
MisterMX, line 126: `features.EnableAlphaManagementPolicies` is a provider feature flag that does not come with the runtime. Changing this wont work.

jwhitcraft: oh i see, i guess i'll go back to the drawing board and see if i can get this provided updated to 1.15.x so we can use the beta feature flag.

MisterMX:
Closing this, since the underlying issue #136 is solved.

PR #138, opened 2024-04-29
feat:Including import credentials for project via secretRef

### LINKED ISSUE #48: Support including import credentials for Project via secretRef

I would like to be able to supply credentials using a Secret for the provider when importing a project from an external URL.

Currently the only way to do this appears to be to include the username/token directly in the `importUrl` parameter. I would like to be able to reference a secret instead.

If this sounds reasonable I'm happy to work on contributing this. Thanks.

frkouokam:

API definition for secretRef has been added in project_types.go.

Retrieving secret and adding it to the importUrl has been implemented in controller/project.go (WIP)

Fixes #48 

I have:

- [ ] Read and followed Crossplane's [contribution process].


apis/projects/v1alpha1/project_types.go
```diff
@@ -148,6 +148,14 @@ type ContainerExpirationPolicyAttributes struct {
 	NameRegex *string `url:"name_regex,omitempty" json:"name_regex,omitempty"`
 }
 
+// Struct representing the secret name and the secret namespace to use when importing a project with secret
+type ImportUrlSecretRef struct {
+	Name        string `json:"secretName"`
+	Namespace   string `json:"secretNamespace"`
+	UsernameKey string `json:"username"`
```
abacus3, line 155: Use defaults here.

pkg/controller/projects/projects/project.go
```diff
@@ -149,6 +158,52 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		secret := &corev1.Secret{}
+
+		err := e.kube.Get(ctx, namespacedName, secret)
+
+		if err != nil {
+			return managed.ExternalCreation{}, errors.New(errNoImportSecret)
+		}
+
+		// Obtain the password from the secret.
+		password := string(secret.Data[keySecretRef.PasswordKey])
```
abacus3, line 182: Check if key is present in Secret struct before doing anything with it.
Also please make sure empty string "" is handled properly (nil vs "").

pkg/controller/projects/projects/project.go
```diff
@@ -149,6 +158,52 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		if password == "" {
+			return managed.ExternalCreation{}, errors.New(errPasswordNotFound)
+		}
+
+		// Obtain the username from the secret.
+		username := string(secret.Data[keySecretRef.UsernameKey])
```
abacus3, line 188: Check if key is present in Secret struct before doing anything with it.
Also please make sure empty string "" is handled properly (nil vs "").

pkg/controller/projects/projects/project.go
```diff
@@ -149,6 +158,52 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 	if !ok {
 		return managed.ExternalCreation{}, errors.New(errNotProject)
 	}
+	// Line 161 -> 206 will be export to an util function
+	keySecretRef := cr.Spec.ForProvider.ImportUrlSecretRef
+	importUrl := cr.Spec.ForProvider.ImportURL
+
+	// User has provided secret for the import
+	if keySecretRef != nil && importUrl != nil {
```
abacus3, line 166: Check also for actual non empty OR enforce minLength of Strings in CRD via kubebuilder annotation. Do both (non empty check AND minLength enforcement) for both parameters ideally.

pkg/controller/projects/projects/project.go
```diff
@@ -149,6 +158,52 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		if username == "" {
+			return managed.ExternalCreation{}, errors.New(errUsernameNotFound)
+		}
+
+		// manipulate url to add the secret. If secret is already in the url, it should be overridden
+		// http://username:password@example.com
+		parsedUrl, err := url.Parse(*importUrl)
+
+		if err != nil {
+			return managed.ExternalCreation{}, errors.New(errUsernameNotFound)
```
abacus3, line 198: Wrong error. If we were not able to parse the initially provided URL, the URL is probably in an incorrect format, as we can not parse it here.
Error with a message, displaying an issue with parsing of the provided URL. Also wrap the provided `err` to provide an actual trace of failure context.

pkg/controller/projects/projects/project.go
```diff
@@ -149,6 +158,52 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		err := e.kube.Get(ctx, namespacedName, secret)
+
+		if err != nil {
+			return managed.ExternalCreation{}, errors.New(errNoImportSecret)
+		}
+
+		// Obtain the password from the secret.
+		password := string(secret.Data[keySecretRef.PasswordKey])
+		if password == "" {
+			return managed.ExternalCreation{}, errors.New(errPasswordNotFound)
```
abacus3, line 184: Wrap the provided err to provide an actual trace of failure context.

pkg/controller/projects/projects/project.go
```diff
@@ -149,6 +158,52 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		// http://username:password@example.com
+		parsedUrl, err := url.Parse(*importUrl)
+
+		if err != nil {
+			return managed.ExternalCreation{}, errors.New(errUsernameNotFound)
+		}
+
+		userInfo := url.UserPassword(username, password)
+
+		parsedUrl.User = userInfo
```
abacus3, line 203: Use the new URL for further processing.

pkg/controller/projects/projects/project.go
```diff
@@ -149,6 +158,52 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+			return managed.ExternalCreation{}, errors.New(errPasswordNotFound)
+		}
+
+		// Obtain the username from the secret.
+		username := string(secret.Data[keySecretRef.UsernameKey])
+		if username == "" {
+			return managed.ExternalCreation{}, errors.New(errUsernameNotFound)
+		}
+
+		// manipulate url to add the secret. If secret is already in the url, it should be overridden
```
abacus3, line 193: We should add this descrption also to the CRD property description, so it is clear that already provided credentials in an encoded URL will be overwritten in case a valid secretRef is present as well.

pkg/controller/projects/projects/project.go
frkouokam: MinLength added for ImportUrl.
keySecretRef is not a string.

apis/projects/v1alpha1/project_types.go
```diff
@@ -148,6 +148,14 @@ type ContainerExpirationPolicyAttributes struct {
 	NameRegex *string `url:"name_regex,omitempty" json:"name_regex,omitempty"`
 }
 
+// Struct representing the secret name and the secret namespace to use when importing a project with secret
+type ImportUrlSecretRef struct {
+	Name        string `json:"secretName"`
+	Namespace   string `json:"secretNamespace"`
+	UsernameKey string `json:"username" default:"username"`
+	PasswordKey string `json:"password" default:"password"`
```
MisterMX, line 156: The `default` tag has no effect. To configure default values for Kubebuilder the comment markers need to be used: https://book.kubebuilder.io/reference/markers/crd-validation

pkg/controller/projects/projects/project.go
```diff
@@ -150,6 +159,57 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, errors.New(errNotProject)
 	}
 
+	keySecretRef := cr.Spec.ForProvider.ImportUrlSecretRef
+	importUrl := cr.Spec.ForProvider.ImportURL
+
+	// User has provided secret for the import
+	secretRefIsNotEmpty := keySecretRef != nil && keySecretRef.Namespace != "" && keySecretRef.Name != ""
+	keysAreNotEmpty := keySecretRef != nil && keySecretRef.PasswordKey != "" && keySecretRef.UsernameKey != ""
+	importUrlIsNotEmpty := importUrl != nil && *importUrl != ""
```
MisterMX, line 168: This can be simplified:
```suggestion
	importUrlIsNotEmpty := ptr.Deref(importUrl, "") != ""
```

pkg/controller/projects/projects/project.go
```diff
@@ -150,6 +159,57 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		if err != nil {
+			return managed.ExternalCreation{}, errors.Wrap(err, errParseUrlFailed)
+		}
+
+		userInfo := url.UserPassword(username, password)
+
+		parsedUrl.User = userInfo
+
+		// Override importUrl with the manipulated URL containing the credentials found in ImportSecretRef.
+		*importUrl = parsedUrl.String()
```
MisterMX, line 210: This whole code segment is put into `Create` which is the wrong place for that.

It should be put into `LateInitialize` which runs during `Observed`and should only update `spec.forProvider.importUrl` if it is nil. 

`Create` itself should only rely on `spec.forProvider.importUrl` and not `importUrlSecretRef`.

package/crds/projects.gitlab.crossplane.io_projects.yaml
```diff
@@ -162,8 +162,30 @@ spec:
...
+                        type: string
+                      secretNamespace:
+                        type: string
+                      username:
+                        type: string
+                    required:
+                    - password
+                    - secretName
+                    - secretNamespace
+                    - username
```
MisterMX, line 187: Are `username` and `password` required values or should the provider support imports from public repos?

MisterMX (changes requested):

package/crds/projects.gitlab.crossplane.io_projects.yaml
frkouokam: username and password are required only if `importUrlSecretRef` is not nil. And yes the provider should also support imports from public repos. @abacus3

MisterMX: OK, thank. I think the name `importUrlSecretRef´ is a bit misleading. Can we change it to something more explicit like `importCredentialsSecretRef`?

MisterMX:
@frkouokam once this PR is ready to be merged can set it to "Ready for Review" and squash your changes into a single commit and sign it off?

PR #144, opened 2024-05-09
feat: SamlGroupLinks

jholt96:

This PR adds SamlGroupLink functionality. This allows users to tie Saml groups to Gitlab groups using the samlGroupSync feature for Gitlab EE. 

https://docs.gitlab.com/ee/user/group/saml_sso/group_sync.html

I have:

=> fails

1. Unit tests have been created to test the controller functionality for observe, connect, create, delete
2. Tested in RKE2 cluster. I was able to create and delete saml group sync's on various groups.


pkg/controller/groups/samlgrouplinks/controller.go
```diff
@@ -0,0 +1,189 @@
...
+	client groups.SamlGroupLinkClient
+}
+
+func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
+	cr, ok := mg.(*v1alpha1.SamlGroupLink)
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errNotSamlGroupLink)
+	}
+
+	samlGroupName := *cr.Spec.ForProvider.Name
```
MisterMX, line 110: External resource names should be stored in the external name annotation are retrieved with `meta.GetExternalName(cr)`.

pkg/controller/groups/samlgrouplinks/controller.go
```diff
@@ -0,0 +1,189 @@
...
+	if cr.Spec.ForProvider.GroupID == nil {
+		return managed.ExternalObservation{}, errors.New(errMissingGroupID)
+	}
+
+	groupLink, res, err := e.client.GetGroupSAMLLink(*cr.Spec.ForProvider.GroupID, samlGroupName)
+	if err != nil{
+		if clients.IsResponseNotFound(res) {
+			return managed.ExternalObservation{}, nil
+		}
+		return managed.ExternalObservation{}, errors.Wrap(err, errGetFailed)
```
MisterMX, line 124: Can be simplied:

```suggestion
		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(err, clients.IsResponseNotFound), errGetFailed)
```

MisterMX (changes requested):

pkg/controller/groups/samlgrouplinks/controller.go
jholt96: So for this 'samlGroupName' variable, its not trying to get the name of the object but the samlGroupName that gets passed to Gitlab. My idea behind having that samlGroupName explicitly defined in the forProvider block, is if users want to assign the same saml group to multiple gitlab groups. Let me know if you think this makes sense! 

e.g:
``` yaml
kind: SamlGroupLink
metadata:
  name: mysamlgroup-group1
spec:
  forProvider:
    groupIdRef:
      name: gitlab-group-1
    name: MySamlGroup
---
kind: SamlGroupLink
metadata:
  name: mysamlgroup-group2
spec:
  forProvider:
    groupIdRef:
      name: gitlab-group-2
    name: MySamlGroup
```


I could also change the json field to be more explicit that it is specific to the samlgroup as maybe 'name' is too generic. The reason I left it as 'name' is to be consistent with the upstream gitlab sdk. 

Could look something like e.g
``` yaml
kind: SamlGroupLink
metadata:
  name: mysamlgroup-group2
spec:
  forProvider:
    groupIdRef:
      name: gitlab-group-2
    samlGroupName: MySamlGroup
```

pkg/controller/groups/samlgrouplinks/controller.go
jholt96: I tried to implement this simplification but it caused errors with the resources.Ignore method. The other controller files in the groups folder use the original method, is it necessary?

pkg/controller/groups/samlgrouplinks/controller.go
jholt96: I looked into the package and understand what you meant! I made the changes and added more unit tests to ensure that it is set.

pkg/controller/groups/samlgrouplinks/controller.go
jholt96: @MisterMX any thought on this?

MisterMX: Can you describe what is not working? [`resource.Ignore`](https://pkg.go.dev/github.com/crossplane/crossplane-runtime/pkg/resource#Ignore) is used in many providers without issues.

jholt96: The issue was the clients.IsResponseNotFound function does not implement the right parameter. expected was error not *gitlab.Response. I reworked the statement to have an IsErrorSamlGroupLinkNotFound function in clients that works. Modified the Unit tests and tested it in my running instance. 

``` golang
	if err != nil {
		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(groups.IsErrorSamlGroupLinkNotFound, err), errGetFailed)
	}
```

jholt96:
Resolved previous lint Issues and ran the golangci-lint local with no errors

MisterMX:
@jholt96 I think there is still an issue with the linter. Can you check again?

jholt96:
> @jholt96 I think there is still an issue with the linter. Can you check again?

Just added the comment that the job was complaining about. Sorry about that. I figured running the tool locally would be the same output.

MisterMX (approved):
LGTM. Thank you very much @jholt96!

PR #154, opened 2024-08-05
feat: deprecate emails_disabled in favor of emails_enabled

### LINKED ISSUE #141: API Deprecations should be handled (right now group -> emails_disabled changed to emails_enabled)
### What happened?

We saw issues concerning the group api, after updating our gitlab instance. In our case, the group api deprecated the `emails_disabled` flag and inverted it as `emails_enabled`. This leads to failing group api requests of the gitlab provider.

For now we used a feature flag to disable the new behaviour. But this won't be the solution for long - I think June / the next major version might be the point were the flag could be removed.

Updating the go gitlab client from 0.86.0 to 0.103.0 would fix the gitlab api side changes, but how do we handle this customer facing? This would lead to breaking changes in the managed resource...

### How can we reproduce it?
Just use the provider with an up to date gitlab instance

Crossplane version: not relevant
Crossplane Provider GitLab version: 0.7.0

dariozachow:

This PR fixes the Issue with Gitlab 16, where emails_disabled has been renamed to emails_enabled  on groups.
This is a breaking change even though It should have been none breaking as discussed here.
But when implementing the changes it felt awkward to work with 2 fields witch are dependend on each other.
This makes a clean representation of the actual Gitlab-API.

The go-gitlab dependency has been updated to the latest version as well.

Fixes #141 

I have:

Tests have been updated to test for emailsEnabled instead of emailsDisabled


dariozachow:
@MisterMX @janwillies

apis/groups/v1alpha1/group_types.go
```diff
@@ -113,7 +113,7 @@ type GroupParameters struct {
 
 	// Disable email notifications.
 	// +optional
-	EmailsDisabled *bool `json:"emailsDisabled,omitempty"`
+	EmailsEnabled *bool `json:"emailsEnabled,omitempty"`
```
MisterMX, line 116: To ensure backwards compatibility we should keep `EmailsDisabled` as a deprecated property and add `EmailsEnabled` as the new recommended option.

Similar to what is described in https://github.com/crossplane-contrib/provider-gitlab/issues/141#issuecomment-2247099937 the field `EmailsEnabled` should be late-initialized during observe (pseudocode):

```go
cr.Spec.ForProvider.EmailsEnabled = LateInitialize(cr.Spec.ForProvider.EmailsEnabled, cr.Spec.ForProvider.EmailsDisabled)
````

And after that `EmailsEnabled` can be used throughout the rest of the controller.

**Review by MisterMX — CHANGES_REQUESTED** [review]
LGTM so far. However, we should always provide backward compatibility if possible to provide a clear migration path for users.

File: apis/groups/v1alpha1/group_types.go [hunk 2]
```diff
@@ -111,10 +111,14 @@ type GroupParameters struct {
 	// +optional
 	SubGroupCreationLevel *SubGroupCreationLevelValue `json:"subgroupCreationLevel,omitempty"`
 
-	// Disable email notifications.
+	// Disable email notifications. Deprecated, use emailsEnabled instead.
 	// +optional
 	EmailsDisabled *bool `json:"emailsDisabled,omitempty"`
```
MisterMX, line 116: Deprecation marker should be in a new line (ref. https://go.dev/blog/godoc):

```suggestion
	// Disable email notifications.
	//
	// Deprecated: Use emailsEnabled instead.
	// +optional
	EmailsDisabled *bool `json:"emailsDisabled,omitempty"`
```

MisterMX (changes requested):

Issue #153, opened 2024-08-01
Allow token from secretKeyRef for the Hook resource


tvandinther:
### What problem are you facing?
I would like to create a Gitlab hook with a token specified in a kubernetes secret. Currently I would have to create a composition with the secret to use patching. This method also displays the token directly on the hook resource.

### How could Crossplane help solve your problem?
Introduce a `spec.forProvider.tokenRef` field for this functionality.

dariozachow:
thats a duplicate for this Issue https://github.com/crossplane-contrib/provider-gitlab/issues/133

tvandinther:
Yes it is, thanks for your contribution.

PR #156, opened 2024-08-21
fix(project/hook)!: hooksecret as reference

### LINKED ISSUE #133: Hook token as secret reference
### What happened?

When using the Hook MR for projects, the secret token which verifies the webhook call is written directly into the MR itself.
Thus having the Secret not in a kubernetes Secret.

It would be better giving a secret ref to the hook and read the secret from that location.

### How can we reproduce it?

Create a Hook with the token field set

Crossplane version: 1.14.2
Crossplane Provider GitLab version: 0.6.1

dariozachow:

This PR changes the Hook Resource to take a secret reference as an input instead of having the token directly in the spec fields.

Fixes #133 and #153 

I have:

New Unit tests have been added to confirm the proper usage of the secret in the hook controller.
Old Unit tests have been updated to use the token as a secret reference


MisterMX (commented):
LGTM so far. Thanks a lot @dariozachow. Could you squash your changes into a single commit?

janwillies (approved):
Thanks a lot @dariozachow! We can squash via GH

janwillies:
well that was a race condition :)

Issue #149, opened 2024-07-01
Pending deletion on gitlab


eperdeme:
### What happened?
I attempted to delete a project via kubectl 
```
kubectl delete -f gitlab-test.yaml
```

```
 Normal   DeletedExternalResource      4m29s (x20 over 7m32s)  managed/project  Successfully requested deletion of external resource
```
The project did not delete and the kubectl never returned retrying over and over to delete the project.

The project cannot be immediatly deleted as our group on Gitlab.com has restrictions that projects go in to a pending state for 7 days.
"Deletion pending. This project will be deleted on 2024-07-08. Repository and other project resources are read-only."

### How can we reproduce it?
Tell Gitlab.com to not immediatly delete projects but to add a pending duration of X days before deleting.

### What environment did it happen in?
Crossplane version: 1.14.0
Crossplane Provider GitLab version: v0.8.0

What can we do about this ? The control plane is just looping unless I guess I modify the finerliser but we can't do this each time? Is there an option in the provider to work around this ?

Issue #121, opened 2024-02-05
Immediate deletion of resources


cristian98149:
### What problem are you facing?
When I delete a project managed by the provider, the project is pending for deletion for a certain retention period. If I force the delete of the resource via kubectl the project is deleted.

### How could Crossplane help solve your problem?
Add a way to immediately delete the project after request the deletion of it.

PR #168, opened 2024-11-24
feat: Add immediate deletion of groups

buehlmann:

On GitLab [Premium](https://about.gitlab.com/pricing/premium/) and [Ultimate](https://about.gitlab.com/pricing/ultimate/), the delete action on a group just marks the group as deleted. By default, the group will be deleted seven days in the future.  To delete the group immediately two rest calls are needed:

1. DELETE request to `/groups/:id` which marks the group in the "pending for deletion" state ([docs](https://docs.gitlab.com/ee/api/groups.html#delete-a-group))
2. DELETE request to `/groups/:id` with attributes `permanently_remove=true` and `full_path` set. The group will be deleted immediate.

This change introduce the optional fields `PermanentlyRemove` and `FullPath` on the GitLab group resource. This allows the user to enforce an immediate deletion of GitLab groups.

The same behavior exist by deleting projects (see #121, #149) and could be implemented in the same way as done here for groups. Before this feature can be implemented for projects, one needs to extend the _xanzy/go-gitlab_ client with `DeleteProjectOptions` along the lines of [DeleteGroupOptions](https://github.com/xanzy/go-gitlab/blob/main/groups.go#L617-L623).

I have:

Provider has been built locally and deployed on a AKS cluster. Deletion of GitLab Groups has been successfully end to end tested with and without PermanentlyRemove and FullPath fields set.


dariozachow:
can you rebase on main?

buehlmann:
@dariozachow MR rebased on master

dariozachow:
Unfortunately this only works for subgroups. Is there some way we can filter for this error ``permanently_remove` option is only available for subgroups` after trying the full deletion once? Or ignore top level groups at all

buehlmann:
I would just skip the newly introduced deletion statement if the group to delete is on the root level. Determine if it's a root group, I would by testing `group.path == group.full_path`. If this condition is true, the group is on the root level. What do you think?

dariozachow:
sounds good to me

buehlmann:
@dariozachow delete logic adjusted as discussed

dariozachow (approved):

Issue #207, opened 2025-06-24
Add option to consider projects in pending deletion as deleted


MisterMX:
### What problem are you facing?

Related to https://github.com/crossplane-contrib/provider-gitlab/issues/149

When a project is deleted in Gitlab Premium or Ultimate that has a deletion duration, the MR won't be deleted immediately and instead be reconciled until the deletion duration is over and the project is actually deleted by Gitlab.

### How could Crossplane help solve your problem?

Add a property to the spec that allows specifying if a project should be considered nonexistent (deleted) by the controller if the project is in a pending deletion state.

PR #212, opened 2025-07-03
feat(project): Add option to treat pending deletion as deleted

MisterMX:

Adds a new property to specify if the controller should treat projects in pending deletion state as deleted.

Fixes https://github.com/crossplane-contrib/provider-gitlab/issues/207

I have:

Manually


dariozachow (changes requested):
changes looking good so far. But can you add a Test for this option?

MisterMX:
@dariozachow I added tests. Can you check again?

PR #240, opened 2025-11-12
fix: use correct property for group accesstoken name

### LINKED ISSUE #191: Name of group access token ignored, using metadata name instead
### What happened?
When creating new GroupAccessTokens, the `spec.forProvider.name` seems to get ignored. Instead, the token which I see in the UI is called like `metadata.name`.

### How can we reproduce it?
Create a group with this configuration:
```
    metadata.name = "dont-call-me-that"
    spec.forProvider.name = "correct-name"
```
And see that the created token visible in the UI is called `dont-call-me-that`.

Crossplane version: 
Crossplane Provider GitLab version: v0.10.3

raweber42:

The `name` property of an group-accesstoken is right now ignored. The `metadata.name` is being used instead. This PR fixes this.

Fixes #191

I have:

- [X] Read and followed Crossplane's [contribution process].
- [X] Run `make reviewable test` to ensure this PR is ready for review.

I added the change and saw (as expected) that the test for it starts failing. So I adjusted the test to reflect the new change, too.

### DIFF
--- a/pkg/cluster/clients/groups/zz_accesstoken.go
+++ b/pkg/cluster/clients/groups/zz_accesstoken.go
@@ -51,7 +51,7 @@ func NewAccessTokenClient(cfg common.Config) AccessTokenClient {
 // GenerateCreateGroupAccessTokenOptions generates project creation options
 func GenerateCreateGroupAccessTokenOptions(name string, p *v1alpha1.AccessTokenParameters) *gitlab.CreateGroupAccessTokenOptions {
 	accesstoken := &gitlab.CreateGroupAccessTokenOptions{
-		Name:   &name,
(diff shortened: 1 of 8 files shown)


markussiebert:
Thanks for the fix!

My thought about this: 

- spec.forProvider.name is a required field; previously it was ignored and metadata.name was sent to GitLab.
- Existing access tokens remain valid; their names are not re‑evaluated because Observe does not compare or late‑initialize the Name field and always reports ResourceUpToDate: true. => Already created tokens will not be marked as needing update (updates are unsupported anyway).
- Only newly created tokens will differ: they will now use the correct spec.forProvider.name.

PR #241, opened 2025-11-12
fix: make groupaccesstokens expirydate nullable

### LINKED ISSUE #162: Allow null expiresAt on group accessToken
GitLab 17.7 introduced the ability to create group access tokens without an expiration date. However, this feature is not yet supported by the Crossplane GitLab provider.

Update the `groups.gitlab.crossplane.io/v1alpha1` CRD to allow the `expiresAt` field in `AccessToken` to accept a null value. While making an API call, `expires_at` is still a required field, so it should be explicitly set to null when no expiration date is desired.

Here is an example API call for creating an access token on a GitLab 17.7 instance:

``` shell
curl --location 'https://gitlab.example.com/api/v4/groups/16/access_tokens' \
--header 'PRIVATE-TOKEN: <YOUR-PRIVATE-TOKEN>' \
--header 'Content-Type: application/json' \
--data '{ "name":"test_token", "scopes":["api", "read_repository"], "expires_at":null, "access_level": 30 }'
``` 

The response looks like this:

``` json
{
    "id": 74,
    "name": "test_token",
    "revoked": false,
    "created_at": "2024-10-10T10:24:47.132Z",
    "scopes": [
        "api",
        "read_repository"
    ],
    "user_id": 1632,
    "last_used_at": null,
    "active": true,
    "expires_at": "2025-10-07",
    "access_level": 30,
    "token": "<THE-GENERATED-TOKEN>"
}
```

raweber42:

Fixes #162":
In order to expose the behavior of the gitlab API, Gitlab Group Access Tokens without `ExpiresAt` should be allowed. They will use a default expiry time (which is also covered by a test).

I have:

Added another test to check for scenarios of `ExpiresAt` being set and not set

### DIFF
--- a/pkg/cluster/clients/groups/zz_accesstoken.go
+++ b/pkg/cluster/clients/groups/zz_accesstoken.go
@@ -20,7 +20,6 @@ package groups
 
 import (
 	"strings"
-	"time"
 
 	gitlab "gitlab.com/gitlab-org/api/client-go"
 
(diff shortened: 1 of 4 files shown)


raweber42:
@markussiebert there was one propterty missing to make this work, could you please have a look at my PR here? https://github.com/crossplane-contrib/provider-gitlab/pull/258

PR #244, opened 2025-11-28
feat: support application settings

### LINKED ISSUE #243: Support Application Settings configuration

I'd like to configure my GitLab instance settings using a dedicated CRD.

As of now for kubernetes deployments I am forced to use the https://gitlab.example.com/api/v4/application/settings endpoint for configuration, which is not that optimal...

I'd like to mimic the behavior of the [GitLab Terraform Provider application settings resource](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/application_settings)

I will be drafting a PR for that shortly, but if someone is faster than me then I'd be glad.

BoxBoxJason:

The goal of this PR is to introduce a way to support application settings configuration for the gitlab instance using the provider. This is greatly inspired from the [terraform gitlab provider application settings resource](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/application_settings).

I recommend using a `squash merge` because I made a lot of commits that would appear weird if they stay as is.

Fixes #243

I have:

What was added:

- A new "instance" api, in addition to the projects and groups, to update instance wide settings (for self hosted instances), it does not work on gitlab.com of course.
- A new "ApplicationSettings" CRD / controller / api to reconcile gitlab instance settings
- two generation function that are used to make type assignments / comparisons / generations from reflection of gitlab client-go types

  ```yaml
  kind: ProviderConfig
  metadata:
    name: gitlab-provider
    namespace: default
  spec:
    credentials:
      source: Secret
      method: PersonalAccessToken
      secretRef:
        namespace: crossplane-system
        name: gitlab-credentials
        key: token
  ```
- I apply the applicationSettings manifest (this is a reduced number of values to not jam the description):
  ```yaml
  apiVersion: instance.gitlab.m.crossplane.io/v1alpha1
  kind: ApplicationSettings
  metadata:
    name: boxboxjason-settings
    namespace: default
  spec:
    providerConfigRef:
      name: gitlab-provider
      kind: ProviderConfig
    forProvider:
      maxArtifactsSize: 222
      silentModeEnabled: true
  ```
  
- I check that parameters are applied and no errors appears.
- I update the gitlab settings manually and confirm that they are reconciled


henrysachs:
This is now included in release 0.15.0

PR #254, opened 2025-12-10
feat: support instance runners

### LINKED ISSUE #249: Support Instance Runner

Using the current CRDs, I can create project & group runner but not instance ones. This is an important feature for self hosted gitlab.

The possibility to create an instance runner should be given, either by using the [GitLab user runner principle](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/user_runner) from terraform, which can either be group or instance level.
Or a dedicated instance runner CRD can be implemented

BoxBoxJason:
The goal of this PR is to add the support of the instance runners, similar to what the terraform provider does.

- Here instead of the single `gitlab_user_runner` handling 3 types of runners (instance, groups, projects) we have 2 differents resources (group, project).
 To avoid introducing breaking changes, I decided to just add a 3rd one (instance).
- I am a little concerned because this is **deviating** from the [official gitlabhq/gitlab terraform provider](https://registry.terraform.io/providers/gitlabhq/gitlab).

During my copy/paste spree, I noticed some typos in the other runners code, so I also fixed them.

I recommend a `squash` merge type to keep a clean history and a single commit for this feature

Closes #249 

I have:

- [x] Run Integration Tests to check that the feature works as expected

  ```yaml
  kind: ProviderConfig
  metadata:
    name: gitlab-provider
    namespace: default
  spec:
    credentials:
      source: Secret
      method: PersonalAccessToken
      secretRef:
        namespace: crossplane-system
        name: gitlab-credentials
        key: token
  ```
- I apply the `examples/instance/runner.yaml`
- I check that parameters are applied and no errors appears.
- I check that I can use the generated token to properly register a runner, I attempt manual revoke and confirm reconciliation
- I update the gitlab settings manually and confirm that they are reconciled


henrysachs (approved):
LGTM

PR #264, opened 2025-12-29
feat: upgrade gitlab client-go to v1 with int64 migration

### LINKED ISSUE #262: [DEPS] Upgrade client-go version (and maybe other deps)

The following dependency: gitlab.com/gitlab-org/api/client-go has released a major version: (v1)
It would be nice to transition to it in order to benefit from latest additions to the gitlab API (I'm thinking of project level external integrations) .

It seems like this upgrade breaks A LOT of things, as many fields have transitionned from int -> int64 between the v0 and the v1 of the package.

The dependencies (at least the gitlab sdk) could be upgraded to latest versions to benefit from new features (and security patches)

### LINKED ISSUE #265: [BUG] e2e-tests CI random timeouts
### What happened?

During the "CI" pipeline, there is a frequent bug in the e2e-tests job that makes it timeout.
It sometimes require to restart the job multiple times (up to 6 times 😱) until it passes (with no code changes).

### How can we reproduce it?

Not really sure how, I tried multiple things but couldn't find out a 100% way to make it happen.
Here is the state of my analysis:
- kubectl wait command are run with a 5 minutes timeout. When the job passes, it ALWAYS does in 2m20s so the problem IS NOT that the timeout is too short.
- I thought it might be because of a conflict on the kubeconfig that could happen (for some magical reasons) when there are multiple instances of the job running simultaneously, but that's also not the case.
- My other far fetched guesses are that the problem comes either from the cache OR the `-j2` arguments that limits CPU too much at boot, causing a stale state

Here is an example of [one that failed](https://github.com/crossplane-contrib/provider-gitlab/actions/runs/20595129720), and [one that succeeded](https://github.com/crossplane-contrib/provider-gitlab/actions/runs/20595298589) right after, with 0 code change

This happens in the GitHub Actions environment.
Crossplane version: 1.18.1
Crossplane Provider GitLab version: any dev version

henrysachs:
## Summary

- Upgrade `gitlab.com/gitlab-org/api/client-go` v0.116.0 → v1.10.0
- Migrate all `int` → `int64` (128 files)
- Fix controller name collisions causing e2e flakiness (#265)

fixes #262
fixes #265

## e2e Flakiness Root Cause (#265)

`controller-runtime v0.19.0+` requires unique controller names. Both namespaced and cluster-scoped controllers registered identical names (e.g., `managed/approvalrule`), causing startup crashes.

**Fix:** Added `cluster.` prefix to cluster-scoped controller names via `hack/generate-cluster-scope.go`.

## Changes

| Area | Changes |
|------|---------|
| Dependencies | client-go v1.10.0, crossplane-runtime v2.1.0, controller-runtime v0.22.0 |
| API Types | `*int` → `*int64` in all ID fields (128 files) |
| Controllers | `cluster.` prefix for cluster-scoped names (24 controllers) |
| Tests | All 55 packages pass |

## Testing

```bash
make reviewable  # ✅
make local-deploy  # ✅ provider starts, no crashes
```

Tested 13 resources against GitLab CE - all reach `READY/SYNCED: True`.

## Breaking Changes

- Integer fields: `*int` → `*int64`
- Cluster controller names: `managed/foo` → `managed/cluster.foo`

YAML users: no changes needed.

Signed-off-by: Henry Sachs <henry.sachs@deutschebahn.com>


markussiebert (approved):
👍 LGTM

henrysachs:
As I think this is functional I will go ahead an merge this as its blocking further feature development

PR #274, opened 2026-02-07
fix: remove omitempty in settings observation

### LINKED ISSUE #273: Application Setting status does not display "false" booleans
### What happened?

I use ApplicationSettings to manage my settings and in the resource status, many boolean parameters are not displayed.

This also seems to affect other fields like "outboundLocalRequestWhitelist".

### How can we reproduce it?

Create an ApplicationSettings resource and check the status

### How can we fix it

This is caused by the "omitempty" tag in the ApplicationSettings API, which makes it incapable of displayed booleans that have a default value (false). Remove the "omitempty" tag in all ApplicationSettingsObservation.

Maybe this should be removed in all fields of observation ?

### Env details

I use provider-gitlab v0.16.0

BoxBoxJason:

This PR updates the ApplicationSettingsObservation generator to get rid of the "omitempty" tags, they are causing users a visualization issue for users who want to read their settings through the ApplicationSettings CR.

Fixes #273 

I have:

Nothing to re run here, fortunately this does not impact code / API surface

I just made sure that all of the fields now properly appear in the CR status

### DIFF
--- a/apis/cluster/instance/v1alpha1/zz_generated_settings_types.go
+++ b/apis/cluster/instance/v1alpha1/zz_generated_settings_types.go
@@ -904,799 +904,799 @@ type ApplicationSettingsParameters struct {
 // This struct was generated from gitlab.Settings struct, with just a rework of the json tags.
 type ApplicationSettingsObservation struct {
 	// +optional
-	ID int64 `json:"id,omitempty"`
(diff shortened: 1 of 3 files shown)


henrysachs (approved):
LGTM

PR #276, opened 2026-02-07
feat: support project importURLSecretRef

### LINKED ISSUE #48: Support including import credentials for Project via secretRef

BoxBoxJason:

After a long time lost in the abyss, Closes #48 

This PR makes it possible to use importUrl from a secret reference in the Project CRD !
There was also a small refactor to put the `UpdateStringFromSecret` in a common section to be reusable (and already reused) in multiple packages.

I have:

  ```yaml
  kind: ProviderConfig
  metadata:
    name: gitlab-provider
    namespace: default
  spec:
    credentials:
      source: Secret
      method: PersonalAccessToken
      secretRef:
        namespace: crossplane-system
        name: gitlab-credentials
        key: token
  ```

(This requires having the source authorized on the gitlab instance)

- I apply the `examples/instance/project.yaml` (with a tweak to use importUrlSecretRef)
- I check that the project is created and no errors appears.
- I delete the gitlab project manually and confirm that they are reconciled


BoxBoxJason:
> can you fix the lint error please? @BoxBoxJason

Hey there, I have rebased the branch and the CI now passes without issues, I was waiting for #284 to reduce the cyclomatic complexity introduced in #281 

Things are functionnal now :smile:

henrysachs (commented):
Hey, nice work on the DeepCopy approach in Observe — that makes sense to me!

I have a question about Create though: the comment says "this is only required for observation, as this is the only method where spec can be updated" — but could you double-check whether `UpdateCriticalAnnotations` after `Create()` returns does a full `client.Update(ctx, managed)` to persist the external-name annotation? If so, I think the mutated `ImportURL` (with the resolved secret value) would end up in the Custom Resource's `spec.forProvider.importUrl` field — visible to anyone with read access to the CR.

A second thing I noticed in Observe: the secret gets resolved into `current` (the DeepCopy), but `isProjectUpToDate` at line 183 reads `&cr.Spec.ForProvider` where `ImportURL` is still nil. Doesn't that mean drift on the import URL would never be detected when using `ImportURLSecretRef`? And wouldn't `ResourceLateInitialized` (`!cmp.Equal(current, &cr.Spec.ForProvider)`) always be true since `current.ImportURL` differs from the original?

Would be great if you could verify — might be that I'm missing something here!

henrysachs (approved):
LGTM

PR #277, opened 2026-02-10
Share projects with groups

### LINKED ISSUE #270: Managed resource for sharing projects with groups
Hi. It would be great if the provider supported sharing projects with groups, as provided by the [Gitlab projects API](https://docs.gitlab.com/api/projects/#share-projects).

I am creating projects using the provider and it works great, but I would like to grant permissions on projects to groups that represent engineering squads, as means less management overhead in my use case.

I can work around this by getting the members of a group and granting them individual access, but the group access approach will scale better for me.

Thanks!

infkf:

Fixes #270
Adds `ProjectShareGroup` instance to work with projects and groups

I have:

- Unit test has been added
- CRUD operations working locally 


henrysachs:
@infkf i approved the workflows now. Probably you need to rebase as we had to update the trivy action... 
But after that I hope every action went green and we're able to merge!

PR #285, opened 2026-02-20
feat: Add MergeTrain options

BigGold1310:

Add support for `MergeTrainsEnabled`, `MergeTrainsSkipTrainAllowed`, and `MergePipelinesEnabled` fields to the GitLab Project resource.

I have:


BigGold1310:
@henrysachs Ready for review.

henrysachs (commented):
Thanks for the PR @BigGold1310  — this looks very close.

I may be missing something, but I think there are still two controller-side gaps:

1. `isProjectUpToDate()` in `pkg/namespaced/controller/projects/projects/project.go` (and the generated mirror in `pkg/cluster/controller/projects/projects/zz_project.go`) still does not compare `mergeTrainsEnabled`, `mergeTrainsSkipTrainAllowed`, or `mergePipelinesEnabled`.
   That means a spec change limited to these fields would still be treated as up to date, so `EditProject` would never be called for that change.

2. `lateInitialize()` in the same controller files does not seem to backfill these new fields into `spec.forProvider` when they are unset.
   So adopted/imported resources, or projects relying on GitLab defaults, would not get these values initialized the same way as nearby optional project settings.

The API / CRD / client wiring looks good to me — it just seems the reconcile path is not fully wired for the new fields yet.

henrysachs (approved):
merged and will implement the missing stuff myself

PR #302, opened 2026-04-12
feat: support admin & global permissions instance serviceaccount

BoxBoxJason:

This PR adds the support of:

- A new "admin" field in the common service account observation, that shows if the service account has gitlab admin permissions.
- A new "admin" field in the instance service account spec. It is used to specify if an instance service account should have admin permissions.
- A new `baselinePermissions` field in the instance service account spec. It is used to specify if a service account should be added to all existing top level groups, with the given permission level.
- Two new observation fields in the instance service account, they are used as a quick "cache" mechanism for reconciliation of the `baselinePermissions`.

Closes #294 

I generated extensive unit tests using AI for the newly added helper functions. Since they heavily rely on "complex" concurrency, they must be properly tested.

I have:

I recommend a "squash" type merge to have a single commit in history.

I have:

  - [x] Creation
  - [x] Update
  - [x] Deletion


AyRickk:
@henrysachs @derbauer97 Up ?

henrysachs (approved):
LGTM

PR #305, opened 2026-04-15
fix: runner token expiration renewal

### LINKED ISSUE #304: Runner Token not renewed on expiration
### What happened?

The GitLab provider runner controllers (Instance, Group, and Project runners) do not detect or handle expired runner tokens. When a runner token expires, the resource is not recreated, and no mechanism exists to refresh or re-obtain a new token.

**Current behavior:**
- `Observe()` method marks the resource as `ResourceExists: true` and `ResourceUpToDate: true` as long as the runner exists in GitLab
- The `TokenExpiresAt` timestamp is explicitly preserved from the previous CR status and never updated from the API response
- Expired tokens remain in the connection secret indefinitely
- Users have no way to trigger token rotation without manually deleting and recreating the runner resource

**Expected behavior:**
- When a token expires (TokenExpiresAt < now), the resource should detect this and either:
  - Mark `ResourceUpToDate: false` to trigger an Update, or
  - Trigger a recreation to obtain a new token with a fresh expiration time
- The token in the connection secret should always be valid

**Affected components:**
- `pkg/namespaced/controller/instance/runners/controller.go`
- `pkg/namespaced/controller/groups/runners/controller.go`
- `pkg/namespaced/controller/projects/runners/controller.go`

### How can we reproduce it?

1. Create a Runner resource (Instance, Group, or Project) with a valid `writeConnectionSecretToRef`
2. Wait for the runner token to expire (or create a runner with a short expiration time)
3. Observe that:
   - The resource remains in `Ready=True` state
   - `status.atProvider.tokenExpiresAt` shows the expired timestamp
   - The connection secret still contains the expired token
   - No reconciliation is triggered to obtain a new token

**Example manifest:**
```yaml
kind: Runner
metadata:
  name: test-runner
spec:
  forProvider:
    groupID: 123
    description: "Test runner"
    paused: false
  writeConnectionSecretToRef:
    name: runner-token
    namespace: default
```

Crossplane Provider GitLab version: [e.g., v0.17.x]

**Additional context:**
The issue is in the `Observe()` method of all three runner controllers. The line:
```go
tokenExpiresAt := cr.Status.AtProvider.CommonRunnerObservation.TokenExpiresAt
cr.Status.AtProvider = runners.Generate[Instance|Project|Group]RunnerObservation(runner)
cr.Status.AtProvider.CommonRunnerObservation.TokenExpiresAt = tokenExpiresAt
```
explicitly preserves the old token expiration time instead of checking if it has expired.

BoxBoxJason:

This PR adds an automatic renewal mechanism on runner token expiration for the project, group & instance runners.

All of the changes are covered by unit tests

Fixes #304 

I have:

I have:

  - [x] Creation
  - [x] Update
  - [x] Deletion


pkg/namespaced/controller/instance/runners/controller.go
```diff
@@ -150,10 +151,22 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
 	cr.Status.AtProvider.CommonRunnerObservation.TokenExpiresAt = tokenExpiresAt
+	cr.Status.AtProvider.CommonRunnerObservation.TokenCreatedAt = tokenCreatedAt
+
+	cr.Status.TokenRenewAt = runners.ComputeNextRunnerTokenRotation(
+		&cr.Status.AtProvider.CommonRunnerObservation,
+		cr.Spec.ForProvider.TokenRenewBeforeDays,
+	)
+
+	if runners.ShouldRotateRunnerToken(&cr.Status.AtProvider.CommonRunnerObservation, cr.Spec.ForProvider.TokenRenewBeforeDays, time.Now().UTC()) {
+		return managed.ExternalObservation{ResourceExists: false}, nil
```
henrysachs: Nice improvement overall. One thing worries me here: when renewal is due, `Observe()` returns `ResourceExists: false` even though the runner still exists in GitLab. That pushes Crossplane down the create path, and `Create()` provisions a brand-new runner instead of rotating the token on the existing one. In practice this looks like it would orphan the old runner, change the external name, and leave cleanup attached only to the newest runner. Since the runner client already exposes `ResetRunnerAuthenticationToken`, would it be safer to keep `ResourceExists: true` here and handle token renewal in `Update()` instead? I think the same issue exists in the group/project + cluster mirrors.

pkg/namespaced/controller/instance/runners/controller.go
```diff
@@ -185,6 +198,8 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 
 	meta.SetExternalName(cr, strconv.FormatInt(runner.ID, 10))
 
+	now := metav1.NewTime(time.Now().UTC())
+	cr.Status.AtProvider.CommonRunnerObservation.TokenCreatedAt = &now
```
henrysachs: Small concern on persistence here: `TokenCreatedAt` / `TokenExpiresAt` are only written during `Create()`, and the GitLab details API does not return them later. If these status mutations are not reliably persisted by the managed reconciler, the next `Observe()` can lose the renewal metadata and the auto-renew path will never trigger consistently. Could we move this onto a status-safe persistence path, or add a reconcile-level test that proves the timestamps survive a full create -> observe cycle?

pkg/namespaced/clients/runners/runner.go
```diff
@@ -56,6 +58,42 @@ func IsErrorRunnerNotFound(err error) bool {
...
+		return nil
+	}
+	var createdAt time.Time
+	if renewBeforeDays == nil {
+		if obs.TokenCreatedAt == nil {
+			return nil
+		}
+		createdAt = obs.TokenCreatedAt.Time
+	}
+	t, ok := common.RotationTime(createdAt, obs.TokenExpiresAt.Time, renewBeforeDays)
```
henrysachs, line 160: Another edge case worth guarding: `tokenRenewBeforeDays` only has a minimum bound today. If a user sets a renewal window larger than the actual token lifetime, `RotationTime()` ends up in the past immediately, so every reconcile treats the token as already due. With the current recreate behavior that would cause perpetual runner churn; even after switching to reset semantics it would still cause endless resets. A validation or runtime guard for `renewAt <= createdAt` would make this safer.

pkg/namespaced/controller/instance/runners/controller_test.go
```diff
@@ -303,6 +305,44 @@ func TestObserve(t *testing.T) {
...
+					withExternalName(extName),
+					withSpec(v1alpha1.RunnerParameters{}),
+					withAtProvider(v1alpha1.RunnerObservation{
+						CommonRunnerObservation: commonv1alpha1.CommonRunnerObservation{
+							ID:             1,
+							TokenExpiresAt: &expiredTokenTime,
+						},
+					}),
+				),
+				result: managed.ExternalObservation{
```
henrysachs, line 339: I think this test is currently locking in the recreate behavior above by expecting `ResourceExists: false` once the token is expired. If the intended design is to keep the existing runner and rotate its token, this probably wants to assert that the resource still exists but is not up to date, and then verify a reset call in `Update()`. Might be worth adjusting these expired-token tests now so they encode the intended renewal semantics rather than the current recreate path.

henrysachs:
@BoxBoxJason left some comments even though it is a draft i would like to have this personally.

BoxBoxJason:
Hey there !

Thanks for this pre review with very good insights. I will implement your suggestions and fixes this week end.

henrysachs:
@BoxBoxJason could you rebase this one too and also remove the draft status? Because i think we're quite close to merging

BoxBoxJason:
Hello !

I just performed the rebase,
I am not that confident to have this merged right now, as I did not have time to test the corrections against my gitlab instance this week end. 

I will have more time this thursday to perform all tests and remove the draft status confidently !

BoxBoxJason:
Turns out you were right !

Observation did not properly persist renewAt / createdAt fields set in Create
After a bit of digging around, I found out that using annotations works a lot better for that purpose.

Tell me what you think about this revised (and this time properly tested) version

BoxBoxJason:
Hey there @henrysachs

Any news about the review / merge ?

I think there is a missconfig with the codecov, all changes added here are fully covered. (except api type declarations which cannot be)

markussiebert (approved):
Approving. Reviewed the runner token auto-renewal end-to-end:

- Rotation uses `ResetRunnerAuthenticationToken` from `Update()` (keeps the runner ID); `Observe()` returns `ResourceExists: true, ResourceUpToDate: false` when due — no recreate/orphaning.
- Token timestamps bridge `Create()` -> `Observe()` via annotations (status is reverted after Create; annotations persist via UpdateCriticalAnnotations) and settle in status on the Update path; I traced the rotation cycle and it settles to a single reset — no storm.
- `renewBeforeDays > lifetime` is guarded (`!renewAt.After(createdAt)`), preventing perpetual resets.
- All four of @henrysachs's review comments are addressed, and tests encode the intended reset-on-Update semantics.

Verified `go build ./...`, full `go test ./...`, and `golangci-lint` green on current master (Go 1.26.4). The only red check is the soft codecov coverage delta.

PR #307, opened 2026-04-16
feat: allow access tokens renewal & description

### LINKED ISSUE #306: Access Tokens not renewed on expiration
### What happened?

Project and Group AccessToken controllers do not renew tokens based on expiration.

Current behavior:
- If GitLab reports a token as revoked, Observe treats it as non-existent and Crossplane can recreate it.
- If a token is expired but not marked revoked by GitLab, Observe still reports ResourceExists true and ResourceUpToDate true.
- ExpiresAt is only used for late initialization of spec fields, not as a re trigger.

As a result, expired access tokens can remain in connection secrets without automatic renewal.

This causes "silent" failures: a token is no longer usable, but is marked as perfectly fine in the CR Status.

### Expected behavior

When an AccessToken is expired, the controller should trigger renewal (for example by reporting ResourceExists false so the reconciler recreates it), similar to how revoked tokens are currently handled.

### Affected controllers

- Project Access Token
- Group Access Token

### How can we reproduce it?

1. Create a ProjectAccessToken or GroupAccessToken without an expiration date, this makes it take the instance default expiration
2. Wait until the token expires.
3. Observe the managed resource:
    - It can still reconcile as existing/up-to-date unless GitLab flags it revoked.
    - No automatic renewal occurs for expiration.

Crossplane Provider GitLab version: 0.17.x

BoxBoxJason:
#### New renewal feature

This PR adds the support of access token renewals for both the group and project access tokens.

- If no "expiresAt" field is specified, the token's expiration date is set to 7 days, and  "infinitely" renewed at expiration.
- This also allows updates of the `expiresAt` field

> [!NOTE]
> For context, since GitLab 16.0, the access token resource without a "expiresAt" could NOT be created, it resulted in an error: `POST https://gitlab.example.com/api/v4/groups/13/access_tokens: 400 {message: 400 Bad request - expires_at is missing}`.

#### New description field

A new supported field was added to the resource specs: `description`. It simply allows to set the access token description. This field is only reconciled on token creation, meaning only if the token is manually deleted, or revoked, or expired.

#### Observation refactor

It refactors the observation / late init behaviour of the resource which were not up to repository standard.
- Adds the currently missing `atProvider` filling on observation
- Removes unused late init behaviour.
- Access tokens which have an expired `expiresAt` will no longer be marked as Ready, but will instead trigger error loops (so the user can see that their token is out of date).

All changes are HEAVILY covered by unit tests to make sure everything works, as this is an important change.

Fixes #306 
Closes #295 

I have:

I have:

  - [x] Creation
  - [x] Update
  - [x] Deletion


derbauer97:
@BoxBoxJason Thanks for working on this. 

This is a feature we really need.

I wanted to ask what you think about a small API change around renewal.

From what I understand, the current PR falls back to a default 7-day expiry when `expiresAt` is not set. That solves the GitLab 16.x requirement, but it still gives little control over token lifetime and rotation timing.

Would you be open to a model more like cert-manager handles certificate renewal, with:
- `forProvider.expiresAt` for a fixed expiry
- `forProvider.duration` as an integer number of days
- `forProvider.renewBefore` for the rotation window before expiry, with a reasonable default

The reason I’d keep `duration` in whole days is that GitLab seems to treat token expiry with day precision and rounds it to `00:00 UTC`, so hour-level durations are a bit misleading in practice.

My preference would be:
- exactly one of `expiresAt` or `duration`
- `renewBefore` only valid together with `duration`
- `renewBefore` gets a sensible default if omitted

markussiebert:
## Concerns about the token rotation approach

### 1. Token ID always changes on rotation

I tested this against a GitLab instance. The rotate endpoint **always** creates a new token with a new ID and revokes the old one:

| | Original ID | After Rotate | Old token state |
|---|---|---|---|
| Group token | A | B | revoked=true, active=false |
| Project token | C | D | revoked=true, active=false |

The new ID is not predictable (just the next available ID in the system, not necessarily +1). The comment "Set external name again in case the ID changes after rotation, although it shouldn't" is incorrect — it always changes.

### 2. Setting external name from `Update()` is not persisted by Crossplane

The Crossplane managed reconciler ([reconciler.go](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/reconciler/managed/reconciler.go)) only persists annotations in two cases:
- After `Create()` → via `UpdateCriticalAnnotations()` (retrying annotation updater)
- After `Observe()` with `ResourceLateInitialized=true` → via `r.client.Update()`

After `external.Update()` succeeds, the reconciler **only** calls `r.client.Status().Update()` (status subresource). Annotation changes (including external-name) made during `Update()` are **not written back to the API server**.

This means the rotation flow will be:
1. `Update()` rotates token → new ID (e.g. 200), old ID (e.g. 199) revoked
2. `meta.SetExternalName(cr, "200")` → in-memory only, not persisted
3. Next reconcile reads external-name "199" from API server
4. `Observe()` → GET token 199 → `active=false` → `ResourceExists: false`
5. `Create()` triggered → creates yet another token (ID 201)
6. Token 200 is now orphaned/leaked

Net effect: every "rotation" leaks a token and falls back to delete+create anyway.

Additionally, the project controllers are missing the `meta.SetExternalName` call entirely (only group controllers have it), but given the above, it doesn't matter — neither would work.

### 3. Suggestion: self-descriptive field names for renewal period

Instead of overloading "omit expiresAt = magic 7-day auto-renewal", I'd prefer an explicit field with a self-descriptive name that makes the unit clear:

- `renewalPeriodDays` — clearly communicates: days, not hours/minutes
- `tokenLifetimeDays` — alternative
- `lifetimeDays` — shorter

This avoids ambiguity about what an empty `expiresAt` means and makes it obvious that fractional days or hour-based values aren't supported.

---

What do you think about these points? Happy to discuss alternative approaches for the rotation mechanism — maybe handling it as a delete+create cycle explicitly, or finding a way to persist the new external name (e.g. via late-init or a custom annotation updater) would be more robust.

markussiebert:
## Proposed alternative: rotate via the Create path

After investigating the external name persistence issue further, here's an approach that works cleanly within Crossplane's reconciler design.

### The idea

Instead of rotating in `Update()`, use `Observe()` + `Create()`:

1. **`Observe()`**: if the token is inactive, expired, or approaching expiry → return `ResourceExists: false`
2. **`Create()`**: check if an external name already exists (i.e. we had a previous token). If so, call the **rotate** endpoint instead of creating from scratch. If rotate fails (old token already dead), fall back to a fresh create.

```go
func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
    // ...
    if !at.Active || tokenNeedsRenewal(at, cr.Spec.ForProvider) {
        return managed.ExternalObservation{ResourceExists: false}, nil
    }
    // ...
}

func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.ExternalCreation, error) {
    // ...
    existingID := meta.GetExternalName(cr)

    if existingID != "" && existingID != cr.GetName() {
        // Previous token exists → rotate (atomic: revokes old + creates new)
        id, _ := strconv.Atoi(existingID)
        at, _, err := e.client.RotateGroupAccessToken(groupID, int64(id), rotateOpts)
        if err != nil {
            // Old token already expired/revoked → create fresh
            at, _, err = e.client.CreateGroupAccessToken(groupID, createOpts)
            if err != nil {
                return managed.ExternalCreation{}, errors.Wrap(err, errCreateFailed)
            }
        }
    } else {
        // First time → normal create
        at, _, err = e.client.CreateGroupAccessToken(groupID, createOpts)
    }

    meta.SetExternalName(cr, strconv.FormatInt(at.ID, 10))
    return managed.ExternalCreation{
        ConnectionDetails: managed.ConnectionDetails{"token": []byte(at.Token)},
    }, nil
}
```

### Why this works

- **External name persistence**: The `Create()` path in Crossplane's reconciler calls `UpdateCriticalAnnotations()` — a retrying updater that guarantees the new external name annotation is persisted. This is the only path designed for external name changes.
- **No leaked tokens**: The rotate endpoint atomically revokes the old token and creates the new one in a single API call. No window where both exist or neither exists.
- **Fallback**: If the old token is already dead (expired, manually revoked), rotate will fail and we fall back to a fresh create. Self-healing.
- **`Update()` stays a no-op** (or handles future non-rotation changes). No need to fight the reconciler's design.

### Remaining open: field naming

As mentioned in my previous comment, I'd still prefer a self-descriptive field name for the renewal trigger (e.g. `renewalPeriodDays`, `tokenLifetimeDays`) rather than overloading the semantics of an empty `expiresAt`.

---

> **Note:** This investigation and the proposed approach were developed with AI assistance (brainstorming, Crossplane reconciler code analysis, GitLab API testing). The conclusions should be independently verified and challenged.

BoxBoxJason:
Hey there !

Thanks to both of you for these suggestions and good insight !

### Leaked tokens issue

For the Update function which did not update the "externalName", I have to admit I failed to detect that when I made this PR.
I have recently found a simple way to "fix" the `save new external name on Update()` issue in another crossplane provider project, I merged it 4 days ago: https://github.com/crossplane-contrib/provider-sonarqube/commit/d5c4ae31076a9280bb5a6803612c287ea5a7d62e#diff-9a05e172ee91acc0568ccab7c2326ac4814877222b5bd035f59f6463aa145e94R307
Simply using `e.kube.Update(ctx, cr)`

It would be functionnal here but I do believe @markussiebert  approach is most optimal and "clean". And this allows us to keep the no-op Update; so this is what I will go with.

### Configurable renewal period

You are both totally right, a "magic 7 days" auto renewal is not the best path, the user should be able to configure that.

For this I will use the `RenewalPeriodDays` option, having two different parameters for renewal is an additional difficulty for users, and after trying to implement it, it is not really compatible with the "clean", not leaked approach, this requires extra validation logic, for a "low value" option to my eyes. The token will at most be rotated 59 seconds (default reconciler loop time) after it is expired with our approach, which is acceptable.

BoxBoxJason:
Alright I just included the suggestions and refactored this PR,

I re tested token creation, token revokation using the UI (does get re created), token deletion using the crossplane resource. The update is no back to a no-op.

Tell me what you think about this revised version

dariozachow (approved):

PR #309, opened 2026-05-12
feat: renew managed access tokens before expiry

derbauer97:
Use a two-thirds lifetime threshold for renewal-managed tokens so they rotate before expiry without requiring a low poll interval.

This follow-up updates `renewalPeriodDays` access token rotation to happen after 2/3 of the token's actual lifetime has elapsed, based on GitLab's observed `createdAt` and `expiresAt`, instead of using a fixed advance window. I let myself inspired by cert-manager and how they are doing this for certificates. I decided against another property on the AccessToken itself so we do not have to check for weird configurations.

This keeps renewal-managed tokens rotating before expiry while avoiding the need for a very low poll interval. Our Crossplane user is one of the top users in our 8k users company instance and primarly beacause of polling AccessToken every 60s which is necessary to rotate them fast after expiry. It also helps prevent unnecessary reminder emails to users telling them to rotate tokens manually even though the provider is managing rotation automatically.

The change only applies to `renewalPeriodDays` tokens. Tokens configured with a fixed `expiresAt` keep their current behavior and are not rotated early.

Fixes #

I have:

- Ran focused unit tests for shared access token renewal logic, namespaced clients/controllers, and generated cluster clients/controllers:
  - `go test ./pkg/common ./pkg/namespaced/clients/groups ./pkg/namespaced/clients/projects ./pkg/namespaced/controller/groups/accesstokens ./pkg/namespaced/controller/projects/accesstokens ./pkg/cluster/clients/groups ./pkg/cluster/clients/projects ./pkg/cluster/controller/groups/accesstokens ./pkg/cluster/controller/projects/accesstokens`
- Validated against a local `kind` cluster with the provider running locally.
- Validated against `git-test.tech.rz.db.de` group `516947`:
  - confirmed an active `renewalPeriodDays` token before the threshold did not rotate
  - confirmed an inactive `renewalPeriodDays` token still rotated successfully
- I did not run `make reviewable test`.


markussiebert (commented):
Nice change overall — one suggestion:

**`time.Now()` coupling in `ShouldRotateToken`**

`HasReachedRenewalTime` already accepts `now` as a parameter (fully testable), but `ShouldRotateToken` hardcodes `time.Now().UTC()`. Consider threading `now` through as a parameter:

```go
func ShouldRotateToken(active bool, createdAt *time.Time, actualExpiresAt *gitlab.ISOTime, desiredExpiresAt *time.Time, now time.Time) bool {
```

Callers pass `time.Now().UTC()`:
```go
return common.ShouldRotateToken(a.Active, createdAt, a.ExpiresAt, desiredExpiresAt, time.Now().UTC())
```

This makes the entire chain deterministic in tests without relying on extreme date ranges (2020→2099). Same pattern already used in `HasReachedRenewalTime`.

Not blocking — the current approach works, but this would make future test cases simpler and less fragile.

markussiebert:
**Request: surface scheduled rotation time in status**

For observability and alerting, it would be very helpful to expose a `renewAt` field in `AccessTokenObservation`:

```go
// RenewAt is the time at which the provider will rotate this token.
// Only populated when renewalPeriodDays is set.
RenewAt *metav1.Time `json:"renewAt,omitempty"`
```

Populated during Observe using `RenewalTime(createdAt, expiresAt)`.

This enables:
- `kubectl get accesstoken` showing when rotation is scheduled
- Prometheus alerts on tokens where `renewAt` is approaching but rotation hasn't happened (e.g. provider is down or GitLab API is failing)
- General visibility into the rotation lifecycle

derbauer97:
image

I added the features and it looks good to me

markussiebert (approved):
LGTM — clean implementation. time.Now() decoupled, renewBeforeDays override with proper CEL validation, and status.renewAt for observability. Ship it.

PR #368, opened 2026-06-24
feat(instance): add ServiceAccountAccessToken managed resource

markussiebert:
## Description

Adds an instance-scoped `ServiceAccountAccessToken` managed resource that manages
the personal access token of a GitLab **instance** service account, mirroring the
group `ServiceAccountAccessToken` resource added in #326.

Instance service accounts have no dedicated token endpoints, so the token is
managed through the personal access tokens API. The resource is keyed by
`serviceAccountId` (the service-account user ID) and supports two modes, selected
automatically:

- **Owner mode (default):** the ProviderConfig authenticates with an instance-admin
  token; the managed token is created/observed/rotated/revoked via the admin
  user-token endpoints.
- **Self mode:** the ProviderConfig authenticates with the very token this resource
  manages (self-sustaining rotation loop); the provider uses the `self` endpoints
  and needs no admin token once bootstrapped.

Both cluster-scoped and namespaced variants are generated.

## Notes for reviewers

This branch builds on top of #326 (now merged). The same review fixes raised there
have been carried into the instance resource:

- `UseLegacyProviderConfig` populates `CredentialsSecretRef` (self-mode detection
  depends on it) — inherited from master.
- Self-mode safety guard: refuse to manage the token when it does not belong to
  `spec.forProvider.serviceAccountId` (guards against a miswired credentials secret
  rotating/revoking the wrong service account's token).
- Owner-mode rotation only falls back to a fresh create when the existing token is
  genuinely gone (404, or 400 "token already revoked"); any other error surfaces
  rather than silently minting a second token.
- Cluster-scope self-mode detection compares the write reference's namespace
  (`w.Namespace`) instead of `cr.GetNamespace()`, since cluster-scoped resources
  have no namespace.

## How was it tested

- `go build ./...`, `go vet ./...` — clean.
- `go test ./...` — full suite passes.
- New unit tests cover self-mode service-account match/mismatch and the owner-mode
  rotation fall-through (404 and 400-revoked), in both namespaced and cluster scopes.


derbauer97:
## Review + live e2e battle-test

Reviewed the code and ran a full end-to-end test: kind + Crossplane 2.3.3, provider running out-of-cluster against a self-managed **GitLab 18.11.5-ee** instance. Exercised both the namespaced (`instance.gitlab.m.crossplane.io`) and cluster-scoped (`instance.gitlab.crossplane.io`) variants.

Overall this is solid and faithfully mirrors #326, with all four of that PR's review fixes correctly carried over. A couple of findings below.

### Verified working end-to-end (namespaced + cluster-scoped)
- **Owner mode**: create/observe -> `Ready/Synced`, token written to a `connection.crossplane.io/v1alpha1` secret and authenticates as the service account.
- **Self mode**: auto-detection (`SelfManaged=True`), self-rotation (token id changes, new value written back, old token revoked, loop stays `Ready/Synced`).
- **Cluster-scoped self mode**: works -- i.e. the `w.Namespace` comparison + legacy `CredentialsSecretRef` carry from #326 are effective for the instance resource too (this was the path that took several rounds to get right in #326).
- **Self-mode service-account mismatch guard**: refuses with a terminal error, no rotate/revoke.
- **Dead/invalid self credential**: clean terminal error with "reseed the credentials secret" guidance, no rotation thrash.
- **CEL**: immutability of `name`/`scopes`/`serviceAccountId` enforced; `expiresAt` XOR `renewalPeriodDays` and the `renewBeforeDays` dependency rules all enforced.

### Owner-mode `createOwner` rotates the external-name token without an ownership check

`createOwner` rotates whatever id is in `crossplane.io/external-name` via the **unscoped admin** endpoint, before validating it belongs to `serviceAccountId`:
- `pkg/namespaced/controller/instance/serviceaccountaccesstokens/controller.go:324-325`
- `pkg/cluster/controller/instance/serviceaccountaccesstokens/zz_controller.go:326-327`

`observeOwner` (`controller.go:266`) does check ownership and returns `ResourceExists:false` on a `UserID` mismatch -- but for a non-deleting resource that just routes into `Create` -> `createOwner`, which rotates first. So the guard triggers the rotation instead of preventing it.

**Reproduced live:** resource with `serviceAccountId` = a service account, external-name drifted to a token id owned by an **admin user**. The provider rotated the admin's token, wrote the resulting **admin-scoped** token into the tenant's connection secret, and reported `Synced=True` (silent). Because `external-name` is editable by anyone who can edit the resource and token ids are small sequential integers, a namespace tenant referencing an admin owner-mode `ProviderConfig` can harvest a freshly-rotated, valid token for **any** user on the instance, including admins.

The group resource isn't exposed to this because its rotate/revoke endpoints are scoped by `group + serviceAccountId` (a foreign id 404s and safely falls back to create); the instance resource lost that implicit scoping by using `RotatePersonalAccessTokenByID`.

The `Delete` path (`RevokePersonalAccessTokenByID`, `controller.go:409`) has the same unscoped shape but is **not** exploitable: on deletion the observe mismatch yields `ResourceExists:false`, so Crossplane skips `Delete`.

**Suggested fix:** in `createOwner`, only rotate `existingID` when its `UserID == serviceAccountId` (reuse the observe result or re-fetch and compare); otherwise drop the external name and create fresh. This restores the scoping the group resource gets for free.

### Owner mode mints PATs for arbitrary users (no service-account validation)

`serviceAccountId` is just a user id handed to `Users.CreatePersonalAccessToken`; nothing checks it's actually a service account. Confirmed it happily creates a PAT for a human admin. Cheap guards (verified on this instance):
- `GET /users/:id` -> a service account returns `bot:true, external:true`; a human returns both `false`. (`user_type` is unreliable -- `null` even for the SA on 18.11.)
- `GET /service_accounts` lists only service accounts -- authoritative if you want a strict check.

Recommend rejecting non-service-account targets before create.

### Minor
- The self-mode mismatch guard is gated on `serviceAccountId` being set, but the self example in `examples/instance/serviceaccountaccesstoken.yaml` omits it -- the shipped example disables its own safety check. Suggest setting `serviceAccountId` there.
- No `shortName` on the instance CRD (group has `saat`); consider `isaat`.

### Open questions
- For `serviceAccountId` validation, prefer the cheap `bot`/`external` Users-API check or the stricter `/service_accounts` membership check?

markussiebert:
Thanks for the thorough review and the live e2e run, @derbauer97 — both security findings are fixed in b4e8ce2.

**Finding 1 — owner-mode rotation without ownership check (privilege escalation)**

`createOwner` now confirms the external-name token actually belongs to the configured `serviceAccountId` before touching it. It fetches the token via `GetSinglePersonalAccessTokenByID` and only rotates when `UserID == serviceAccountId`; on a mismatch (or 404) it drops the external name and creates a fresh token for the correct service account instead of rotating the foreign one. This restores the scoping the group resource gets implicitly. The non-404 error path still surfaces (no silent second-token minting), as before.

Added a regression test `OwnerForeignExternalNameCreatesFresh`: external-name drifted to a token owned by a different user → the guard refuses to rotate (the rotate mock is nil, so any rotate call panics the test) and a fresh token is minted instead. Plus `OwnerOwnershipCheckGetError` for the non-404 surfacing.

**Finding 2 — minting PATs for arbitrary users**

`createOwner` now rejects non-service-account targets before create. `validateServiceAccount` does `GET /users/:id` and requires `bot && external`, refusing to issue a token otherwise. I went with the cheap Users-API `bot`/`external` check rather than `/service_accounts` membership — it's O(1), avoids pagination, and matches what you verified on the instance (and `user_type` being unreliable). Added `GetUser` to the client interface. Tests: `OwnerCreateRejectsNonServiceAccount` and `OwnerCreateGetUserFailed`.

> For `serviceAccountId` validation, prefer the cheap `bot`/`external` Users-API check or the stricter `/service_accounts` membership check?

Went with `bot`/`external`. Happy to switch to `/service_accounts` if you'd rather have the authoritative membership check. Note that ref-based usage (`serviceAccountIdRef`) always points at a managed `ServiceAccount`, so the check only really bites on a hand-coded raw id pointing at a human.

**Minor**

- Set `serviceAccountId` on the self-rotating example so its self-mode mismatch guard is active.
- Added `shortName: isaat` to the instance CRD.

Cluster-scoped `zz_` controller/client/tests and the CRDs were regenerated, not hand-edited.

dariozachow (approved):

PR #372, opened 2026-06-25
feat: add managed resource for GitLab group webhooks (group Hook)

### LINKED ISSUE #371: Add managed resource for GitLab group webhooks (groups Hook)
## What problem are you facing?

`provider-gitlab` exposes a managed resource for **project**-level webhooks (`Hook` in `apis/.../projects/v1alpha1`, `zz_hook_types.go`), but there is **no equivalent managed resource for group-level webhooks**.

`apis/.../groups/v1alpha1` currently contains: `AccessToken`, `Badge`, `DeployToken`, `Group`, `GroupMember`, `LdapGroupLink`, `Runner`, `SamlGroupLink`, `ServiceAccount`, `ServiceAccountAccessToken`, `Variable` — but no `Hook`/webhook type.

GitLab itself supports group webhooks (Premium+) via the Group Hooks API, and the underlying `xanzy/go-gitlab` client already implements it (`group_hooks.go`, `GroupHooksService`). So the only missing piece is the Crossplane managed resource.

## How could this provider help solve your problem?

Add a `Hook` managed resource under the `groups` API group (mirroring the existing project `Hook` and the `GroupMember` pattern for `group_id` referencing), wired to the GitLab Group Hooks API:

- `POST   /groups/:id/hooks`
- `GET    /groups/:id/hooks/:hook_id`
- `PUT    /groups/:id/hooks/:hook_id`
- `DELETE /groups/:id/hooks/:hook_id`

Suggested `forProvider` fields, matching the Group Hooks API / `go-gitlab` `AddGroupHookOptions`:

- `group` / `groupRef` / `groupSelector` (group id reference, like `GroupMember`)
- `url` (required)
- `token` (secret ref, write-only)
- `pushEvents`, `mergeRequestsEvents`, `pipelineEvents`, `jobEvents`, `noteEvents`, `confidentialNoteEvents`, `issuesEvents`, `confidentialIssuesEvents`, `tagPushEvents`, `wikiPageEvents`, `deploymentEvents`, `releasesEvents`, `subgroupEvents`, `emojiEvents`, `memberEvents`
- `enableSSLVerification`
- `pushEventsBranchFilter`

### Use case

Subscribing to events for an entire GitLab group (and its subgroups) with a single webhook, rather than registering a project hook on every project. This is useful for platform-level integrations that react to merge-request / pipeline / note events across a whole group.

I'm happy to work on a PR for this if the approach sounds reasonable.

markussiebert:
## Description of your changes

Adds a `Hook` managed resource under the `groups` API group, wired to the GitLab Group Hooks API (`POST/GET/PUT/DELETE /groups/:id/hooks`). It mirrors the existing project `Hook` and reuses the group-reference pattern (`group`/`groupRef`/`groupSelector`) from the group `Member` resource.

Supported `forProvider` fields:
- `url` (required)
- `token` (optional, write-only secret ref)
- `groupId` / `groupIdRef` / `groupIdSelector`
- `pushEvents`, `pushEventsBranchFilter`, `issuesEvents`, `confidentialIssuesEvents`, `mergeRequestsEvents`, `tagPushEvents`, `noteEvents`, `confidentialNoteEvents`, `jobEvents`, `pipelineEvents`, `wikiPageEvents`, `deploymentEvents`, `releasesEvents`, `subgroupEvents`, `emojiEvents`, `memberEvents`
- `enableSslVerification`

The token is optional, matching the GitLab API for group webhooks.

Includes the namespaced source, the generated cluster-scoped variant, deepcopy/managed methods, CRDs, client and controller unit tests, and an example manifest.

Fixes #371

## How has this code been tested

`go build ./...`, `go vet ./...`, and `go test ./...` all pass. New unit tests cover the group hook client (late-init, observation, create/edit options, up-to-date) and controller (Observe/Create/Update/Delete), mirroring the project hook tests for both the namespaced and generated cluster-scoped variants.


derbauer97:
## Review note: webhook `token` rotation isn't reconciled (applies to project `Hook` too)

Reviewed the PR — it's a clean mirror of the project `Hook` and `go build`/`go vet`/`go test ./...` pass locally. One behavioral gap worth addressing around the write-only `token`.

### The gap
GitLab never returns the webhook secret on `GET /groups/:id/hooks/:id`, so `IsHookUpToDate` (correctly) doesn't compare it. The side effect:
- The token is only sent inside `Create` and `Update`.
- `Update` only runs when `IsHookUpToDate` returns `false`, i.e. when some *other* field (events/URL/SSL) drifts.
- So if a user rotates **only** the token in the referenced secret, `Observe` keeps returning `ResourceUpToDate: true` and the new secret is **never** pushed to GitLab — the hook silently keeps validating payloads with the old token.

Because GitLab won't echo the token back, we can't detect drift from the external side. But we don't need to — we can detect change on the **desired** side.

### Suggested fix (desired-side fingerprint)
1. In `Create`/`Update`, after sending the token, store a hash of the resolved value (e.g. SHA-256) in `status.atProvider` (e.g. `tokenHash`) — never the raw token.
2. In `Observe`, resolve the secret, hash it, and compare to the stored hash; mismatch (or a previously-set hash that no longer resolves) ⇒ `ResourceUpToDate: false`.
3. That drives `Update`, which pushes the current token and rewrites the stored hash.

Cost: `Observe` must resolve the token secret on every poll (today only `Create`/`Update` do). Caveats to keep in mind: this reconciles *intent* changes (rotating the K8s secret) but still can't catch someone editing the token directly in the GitLab UI — that's fundamentally unknowable for a write-only field, and that's fine. Storing only a hash in status keeps the secret out of the API object.

### Please apply the same to the project `Hook`
`pkg/namespaced/controller/projects/hooks` (and its generated cluster variant) has the identical set-once behavior. To keep the two resources consistent, the same token-hash handling should be retrofitted there rather than leaving group and project hooks behaving differently.

Happy to grill the rest of the implementation (groupId immutability, not-found handling, delete-by-id) separately if useful.

derbauer97:
## Follow-up: two more items + e2e results

Ran the full lifecycle on a kind cluster (provider out-of-cluster) against a self-managed **GitLab 18.11.5-ee** group: create/observe/update/delete all work — hook created with the right url/events/SSL, `Update` propagated a changed event, and `Delete` removed it (`GET` → 404 afterwards). The webhook `token` is handled cleanly: its raw value never appears in the CR (status holds only `{id, createdAt}`).

Two things worth fixing (both also apply to the project `Hook`):

### 1. `groupId` immutability isn't enforced
`groupId` only carries a `+immutable` doc marker; the generated CRD has no `x-kubernetes-validations`. I confirmed live that `kubectl patch ... groupId: 999999` is **accepted**. If that change sticks, `Observe` looks up the existing hook id under the new group → 404 → `ResourceExists:false` → `Create` makes a new hook in the new group and **orphans** the original. Recommend the same CEL rule the service-account resources use:
```go
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="groupId is immutable"
```

### 2. Drop the dead not-found string matcher
`Observe` has two not-found mechanisms: `clients.IsResponseNotFound(res)` (status code — works) and `resource.Ignore(IsErrorHookNotFound, err)` matching `errHookNotFound = "404 Not found"`. GitLab actually returns `"404 Not Found"` (capital “F”), so the string matcher never fires — it's dead code that only adds confusion. Recommend removing `errHookNotFound`/`IsErrorHookNotFound` and relying on `IsResponseNotFound`.

### Reminder on the token (from the earlier comment)
The e2e also reproduced the token-rotation gap: rotating **only** the referenced secret left the resource `Ready/Synced` with no `Update` triggered — the new secret never reaches GitLab. The hash-in-status approach from the previous comment fixes this. As noted, the project `Hook` shares all three behaviors, so it'd be good to fix both together.

markussiebert:
@derbauer97 addressed your review:

- **`groupId` immutability** — fixed in bab3f01: added the `XValidation:rule="self == oldSelf"` CEL rule (same as the service-account resources) and regenerated both CRDs + the cluster type.
- **Dead not-found matcher** — fixed in 9775b47: removed `errHookNotFound` / `IsErrorHookNotFound` and rely on `clients.IsResponseNotFound` (status-code based). The string matcher was unreachable anyway.
- **Token rotation not reconciled** — split out to #374, since it also affects the project `Hook` and should be solved consistently for both (desired-side hash-in-status).

I also recorded the two project-`Hook` consistency items on #374 (`projectId` immutability and the same dead-matcher removal), to keep this PR scoped to group webhooks.

dariozachow (approved):

PR #403, opened 2026-08-28
feat(serviceaccount): support baseline permissions filters

### LINKED ISSUE #402: Add filters to baseline permissions for instance service accounts

`spec.forProvider.baselinePermissions` on the instance `ServiceAccount` is all or nothing: as soon as it is set, the service account is added to **every** top level group of the instance.

On our instance only a subset of the top level groups is managed by the platform team. The rest belongs to other teams, and we have no business adding a service account to them. Today the only way to avoid that is to not use `baselinePermissions` at all and manage every membership by hand.

Add an optional `spec.forProvider.baselinePermissionsFilters` field, a list of regular expressions matched against the path of the top level groups. The baseline access level would only be granted on a group if at least one expression matches its path.

```yaml
forProvider:
  baselinePermissions: developer
  baselinePermissionsFilters:
    - ^platform-
    - ^infra$
```

An empty or omitted list would keep the current behaviour (all top level groups), so this stays backward compatible.

BoxBoxJason:

Adds an optional `spec.forProvider.baselinePermissionsFilters` field to the instance `ServiceAccount` resource: a list of regular expressions matched against the path of the top level groups. The baseline access level is only granted on a group if at least one expression matches its path.

```yaml
forProvider:
  baselinePermissions: developer
  baselinePermissionsFilters:
    - ^platform-
    - ^infra$
```

> [!NOTE]
> - An empty or omitted list keeps the existing behaviour, every top level group is selected, so this is backward compatible.
> - The expressions are compiled in `Observe`, an invalid one surfaces as a reconcile error instead of being silently ignored.
> - Filtered out groups are skipped before the membership lookup, so this also saves API calls on instances with many top level groups.
> - The baseline permissions part of `Observe` moved to a helper to keep the function under the cyclomatic complexity limit, no behaviour change there.

Closes #402 

I have:

- AI generated unit tests are passing
- Integration (usage) tests against my gitlab instance are passing as well, service account is only added to matching groups


henrysachs (approved):
Reviewed against #402. Cluster/namespaced parity, regex handling, status, filtering, generated CRDs, and tests look correct.

PR #2, opened 2020-12-11
Implement projects api

mscholtens:
### Description of your changes

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

- Run all unit tests with success
- Successful manual testing

[contribution process]: https://git.io/fj2m9

PR #405, opened 2026-08-28
fix(protectedbranch): preserve UserID/GroupID and fix reconcile thrash for identity-only access rules

### LINKED ISSUE #267: ProtectedBranch ignores UserID and GroupID in access levels
## Description

The `ProtectedBranch` resource currently ignores `UserID` and `GroupID` fields in `PushAccessLevels`, `MergeAccessLevels`, and `UnprotectAccessLevels`, only using the `AccessLevel` field. This prevents users from configuring granular, user-specific or group-specific branch protection rules that GitLab's API supports.

## Current Behavior

When creating a ProtectedBranch with user or group-specific access:

```yaml
apiVersion: projects.gitlab.m.crossplane.io/v1alpha1
kind: ProtectedBranch
metadata:
  name: example-protected-branch
spec:
  forProvider:
    projectId: "123"
    branchName: "main"
    pushAccessLevels:
      - accessLevel: 40  # Maintainer level
      - userId: 456      # Specific user
      - groupId: 789     # Specific group
    mergeAccessLevels:
      - accessLevel: 30  # Developer level
      - groupId: 789     # Specific group
```

**Result:** Only the `accessLevel` is applied; `userId` and `groupId` are silently ignored.

## Expected Behavior

All access level entries (including user-specific and group-specific rules) should be applied to the protected branch, allowing for granular access control as supported by GitLab's API.

## Root Cause

In `pkg/namespaced/clients/projects/protectedbranch.go` (lines 174-188), the `GenerateProtectRepositoryBranchesOptions` function only extracts the **first element** from each access level array and only uses the `AccessLevel` field:

```go
if len(p.PushAccessLevels) > 0 && p.PushAccessLevels[0].AccessLevel != nil {
    accessLevel := gitlab.AccessLevelValue(*p.PushAccessLevels[0].AccessLevel)
    opt.PushAccessLevel = &accessLevel  // ← Only uses AccessLevel from first element
}
// UserID and GroupID are never used
```

The same pattern applies to `MergeAccessLevels` and `UnprotectAccessLevels`.

## Use Cases Being Blocked

1. **CI/CD Bot Access**: Allow only a specific CI/CD bot user to push to protected branches
2. **Release Team**: Restrict push access to a specific "Release Team" group
3. **Mixed Permissions**: Combine role-based, user-specific, and group-specific access rules
4. **Security Teams**: Allow only security team members to modify certain branches

## Proposed Solution

Update the client code to properly handle all elements in the access level arrays and include `UserID` and `GroupID`:

```go
// Pseudocode - proper implementation would use GitLab client-go's actual structs
opt.AllowedToPush = []gitlab.BranchAccessDescription{}
for _, level := range p.PushAccessLevels {
    access := gitlab.BranchAccessDescription{}
    if level.AccessLevel != nil {
        access.AccessLevel = level.AccessLevel
    }
    if level.UserID != nil {
        access.UserID = level.UserID
    }
    if level.GroupID != nil {
        access.GroupID = level.GroupID
    }
    opt.AllowedToPush = append(opt.AllowedToPush, access)
}
// Similar for MergeAccessLevels and UnprotectAccessLevels
```

## Files Affected

- `pkg/namespaced/clients/projects/protectedbranch.go` - Lines 159-190
- `pkg/cluster/clients/projects/zz_protectedbranch.go` - Similar pattern

## Additional Context

The CRD types correctly define these fields as `*int64`:
- `apis/namespaced/projects/v1alpha1/protectedbranch_types.go:42` - `UserID *int64`
- `apis/namespaced/projects/v1alpha1/protectedbranch_types.go:46` - `GroupID *int64`

This is a pre-existing limitation, not introduced by recent changes. The types are correct; only the client implementation needs updating.

## References

- [GitLab API - Protected Branches](https://docs.gitlab.com/ee/api/protected_branches.html)
- GitLab client-go library supports these fields in its API structs

1872etd7q23wd:

This fixes two related bugs in `ProtectedBranch` that together caused the delete/recreate ("thrash") loop described in #267.

**1. `GenerateProtectRepositoryBranchesOptions` dropped `UserID`/`GroupID`.**
It only ever read the first entry of `PushAccessLevels`/`MergeAccessLevels`/`UnprotectAccessLevels` and mapped its `AccessLevel` onto the deprecated singular push/merge/unprotect access level fields. `UserID`- and `GroupID`-scoped entries, and any entry beyond the first, were silently dropped from the API request GitLab actually received. Fixed by sending the array-based `allowed_to_push`/`allowed_to_merge`/`allowed_to_unprotect` options instead, preserving `AccessLevel`, `UserID`, and `GroupID` for every entry.

**2. `isAccessLevelsUpToDate` required `AccessLevel` to match even when the spec didn't set one.**
For the common case of a userId-only or groupId-only rule (e.g. `pushAccessLevels: [{userId: X}]`, no `accessLevel`), the match check only ran when `specLevel.AccessLevel != nil`. Since that's never true for an identity-only rule, `isAccessLevelsUpToDate` returned `false` unconditionally — meaning even after fix #1 makes the provider send/observe the userId correctly, `IsProtectedBranchUpToDate` would still report "not up to date" on every single reconcile. `Update()` has no real PATCH for protected branches — it always unprotects and re-protects the branch (see `controller.go`) — so this alone reproduces the exact delete/recreate thrash from the issue. Fixed by only comparing `AccessLevel` when the spec actually sets it.

I verified fix #2 was necessary, not theoretical, by testing directly against a real GitLab (Premium-tier) project: sending `allowed_to_push: [{"user_id": X}]` is accepted and persisted correctly, and GitLab reports it back as `{"access_level": 40, "user_id": X, ...}` — a real, non-zero `access_level` alongside the `user_id`. Without fix #2, that response would never satisfy the old comparison for a spec with no `accessLevel` set, regardless of fix #1.

Fixes #267

I have:

- **Live verification against real GitLab** (Premium-tier project, created and torn down for this purpose):
  - Confirmed the array-based `allowed_to_push: [{"user_id": X}]` request (what fix #1 now sends) is accepted by GitLab's API and round-trips correctly on read-back, for both a project owner/member and a service-account user.
  - Confirmed GitLab's response includes a real `access_level` value alongside the `user_id` (not zero/absent) — which is exactly what makes the old `isAccessLevelsUpToDate` (pre-fix #2) permanently return "not up to date" for identity-only rules, since it required `access_level` to match too.
- Added unit tests covering both fixes: a `UserID`-only push rule (request-building), a mixed `AccessLevel`+`UserID` array, a `GroupID`-only merge rule, the no-access-levels case, and `IsProtectedBranchUpToDate` scenarios for userId-only match/mismatch and the existing accessLevel-only behavior — all pass.
- Full existing test suite for the two affected packages (`pkg/namespaced/clients/projects`, `pkg/cluster/clients/projects`) passes with no regressions.
- `go build ./...`, `go vet`, and `gofmt -l` are clean.
- `make reviewable test` (generate + lint + test) passes end-to-end across the whole repository: `make generate` produces no diff (the generated `pkg/cluster/clients/projects/zz_protectedbranch.go`, mirrored from the namespaced client via `hack/generate-cluster-scope.go`, stays in sync); `make lint` (golangci-lint v2.11.2) reports 0 issues; `make test` passes repo-wide.


pkg/namespaced/clients/projects/protectedbranch.go
```diff
@@ -226,15 +247,15 @@ func isAccessLevelsUpToDate(specLevels []*v1alpha1.BranchAccessDescription, gitl
...
-				userMatch := (specLevel.UserID == nil && gitlabLevel.UserID == 0) || (specLevel.UserID != nil && *specLevel.UserID == gitlabLevel.UserID)
-				groupMatch := (specLevel.GroupID == nil && gitlabLevel.GroupID == 0) || (specLevel.GroupID != nil && *specLevel.GroupID == gitlabLevel.GroupID)
-
-				if userMatch && groupMatch {
-					found = true
-					break
-				}
+			// AccessLevel is only compared when the spec sets it; a userId/groupId-only
+			// rule must match on identity alone, or it can never be considered up to date.
+			accessMatch := specLevel.AccessLevel == nil || int64(*specLevel.AccessLevel) == int64(gitlabLevel.AccessLevel)
```
henrysachs: Each desired entry currently scans the full observed slice, so duplicate desired rules can reuse one observed rule; nil desired entries also panic here, unlike `branchPermissionOptions`. Filter nil desired entries before the cardinality check, then record each matched observed index.

```suggestion
	filteredSpecLevels := make([]*v1alpha1.BranchAccessDescription, 0, len(specLevels))
	for _, specLevel := range specLevels {
		if specLevel != nil {
			filteredSpecLevels = append(filteredSpecLevels, specLevel)
		}
	}
	if len(filteredSpecLevels) != len(gitlabLevels) {
		return false
	}

	matched := make([]bool, len(gitlabLevels))
	for _, specLevel := range filteredSpecLevels {
		found := false
		for i, gitlabLevel := range gitlabLevels {
			if matched[i] || gitlabLevel == nil {
				continue
			}

			accessMatch := specLevel.AccessLevel == nil || int64(*specLevel.AccessLevel) == int64(gitlabLevel.AccessLevel)
			userMatch := (specLevel.UserID == nil && gitlabLevel.UserID == 0) || (specLevel.UserID != nil && *specLevel.UserID == gitlabLevel.UserID)
			groupMatch := (specLevel.GroupID == nil && gitlabLevel.GroupID == 0) || (specLevel.GroupID != nil && *specLevel.GroupID == gitlabLevel.GroupID)
			if accessMatch && userMatch && groupMatch {
				matched[i] = true
				found = true
				break
			}
		}
		if !found {
			return false
		}
	}

	return true
```

henrysachs (changes requested):
Please add regression coverage for duplicate/cardinality and nil desired entries, and regenerate the cluster mirror from the namespaced source to keep generated parity.

henrysachs:
also just a tip, your username seems a bit suspicious for supply chain attacks so maybe you want to make it a bit clearer who you are :)

pkg/namespaced/clients/projects/protectedbranch.go
1872etd7q23wd: Applied this and found one more issue while adding the requested tests: the matching is still greedy. Each desired entry takes the first compatible observed slot and keeps it which is fine for exact matches but an unconstrained entry (no AccessLevel/UserID/GroupID set) is compatible with any role-based observed entry. It can grab a slot that a more specific entry needed, making the result depend on entry order:

```
specA := &v1alpha1.BranchAccessDescription{}                      // no constraints
specB := &v1alpha1.BranchAccessDescription{AccessLevel: &level30}
gitlabLevels := []*gitlab.BranchAccessDescription{{AccessLevel: 30}, {AccessLevel: 40}}
```

```
isAccessLevelsUpToDate([]*v1alpha1.BranchAccessDescription{specA, specB}, gitlabLevels) // false
isAccessLevelsUpToDate([]*v1alpha1.BranchAccessDescription{specB, specA}, gitlabLevels) // true
```

Same state but different answer  specA takes the level30 slot even though it could've taken 40 and leaves specB with no match. Since Update() always unprotects/re-protects on "not up to date," this reintroduces the same thrash bug behind a narrower trigger.

Replaced the greedy assignment with augmenting-path bipartite matching (tryAssign), added a test asserting both orderings agree, and re-ran make generate/make reviewable test clean.

@henrysachs  take a look and let me know if you agree with this fix

henrysachs:
@1872etd7q23wd please merge the latest master into your branch so the trivy finding is fixed. Else the pipeline won't let us merge 😅

henrysachs (approved):

henrysachs:
I just learned I can merge anyway. NICE!

PR #408, opened 2026-09-04
feat(providerconfig): add namespaced v1beta2 credentials surface

markussiebert:
## Summary

Adds `gitlab.m.crossplane.io/v1beta2` `ProviderConfig` as the storage version of
the namespaced provider config, with a credentials surface that fits the
namespaced scope:

- `secretRef` is a `LocalSecretKeySelector`, always resolved in the
  ProviderConfig's own namespace (no `namespace` field);
- the filesystem and environment sources are not offered — they reference
  provider-global state (the provider pod's filesystem / process environment)
  and cannot be scoped to a single namespace without breaking isolation;
- the credentials `source` enum is narrowed to `None;Secret`.

`v1beta1` is retained (served, non-storage) and only gains a version-level
deprecation warning steering users to `v1beta2`. `ClusterProviderConfig` and
`ProviderConfigUsage` are unchanged (v1beta1 only) — they legitimately need a
namespaced secret selector / have no successor.

The provider resolves namespaced credentials via the v1beta2 ProviderConfig and
records the resolved namespace in `Config.CredentialsSecretRef`, so self-managed
token detection keeps comparing against the managed resource namespace.

## Conversion

Strategy `None` (no webhook). The v1beta1 -> v1beta2 delta is only
dropped/optional fields and a narrowed enum, none of which needs data
transformation. Because v1beta2 is storage, fields it does not define
(`secretRef.namespace`, `fs`, `env`) are pruned when a v1beta1 object is written,
and the provider only ever reads v1beta2 — so namespaced credentials can no
longer be sourced across namespaces regardless of the version applied.

## Backward compatibility

- v1beta1 schema is unchanged except the deprecation marker; existing v1beta1
  manifests and stored objects keep working.
- `ClusterProviderConfig` / `ProviderConfigUsage` CRDs are byte-identical to
  before.
- No functional regression versus current master: `secretRef.namespace` was
  already ignored (own-namespace enforced) and `fs`/`env` were never
  implemented. The only visible change is that these already-ignored fields no
  longer persist on v1beta1 (pruned at v1beta2 storage), plus the deprecation
  warning.
- Dropping v1beta1 in the future will require a storage-version migration; it is
  intentionally kept served here so nothing must migrate now.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` pass; `make generate`
  produces no drift.
- Live end-to-end against a real GitLab instance via a kwok cluster with the
  provider running out-of-cluster:
  - native v1beta2 owner-mode token create -> Ready;
  - v1beta2 self-managed detection (`SelfManaged=True`) and one forced
    self-rotation (token id changes, then settles, no rotation storm);
  - v1beta1 manifest -> None conversion -> Ready (deprecated `namespace` pruned
    at storage);
  - the v1beta1 deprecation warning is emitted on `kubectl apply`.


dariozachow (approved):

dariozachow (approved):

henrysachs:
Thanks @markussiebert — this addressed everything from the last pass, and in a couple of places more cleanly than I suggested. 

LGTM

henrysachs (approved):

PR #411, opened 2026-09-18
feat(groups,projects): accept the full path as external-name

### LINKED ISSUE #318: Support GitLab group resource with full path
I want to create a observe-only GitLab group resource, but I only know the full path of the group, but not the ID.

Similar to the GitLab API and the terraform provider for GitLab, support lookup via the full path and not only the ID.

Example:
```
kind: Group
metadata:
  name: team-a-observed
  annotations:
    crossplane.io/external-name: platform/team-a  # At the moment only a numeric ID is supported here
spec:
  managementPolicies:
    - Observe
```

ahauserv:

I want to declare observe-only `Group` and `Project` resources for cases where only the full path is known, not the numeric ID.

- A numeric `crossplane.io/external-name` is still the ID.
- A non-numeric one is sent to the lookup as the full path. Update and Delete already passed the annotation through unchanged, so only the check in `Observe` is removed.
- The annotation is not rewritten to the ID. A Composition re-applies the annotation on every reconcile, so any value the provider wrote would be reverted and, with `LateInitialize`, loop.

> [!NOTE]
> GitLab follows its redirect routes for renamed projects, so a project lookup by the old path keeps working. It does not for groups; an observe-only group renamed in GitLab needs its annotation updated. A fully managed group is renamed back by drift correction anyway.

Fixes #318

I have:

- Unit tests for both controllers cover the path lookup and assert the annotation is left as set.
- Verified against GitLab 19.3.2-ee with a scratch group and project that both are found by full path, and that the old path still resolves for a renamed project but not for a renamed group.


examples/groups/group-observe-only.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: example-group-observed
+  annotations:
+    # Numeric ID or full path of the group.
+    crossplane.io/external-name: path/to/example-group
+spec:
+  managementPolicies:
+    - Observe
+  forProvider:
+    path: example-group
```
tenstad:
```suggestion
  forProvider: {}
```

is `path` redundant here?

tenstad: It's required in the api spec, but I guess it won't really do anything with `crossplane.io/external-name: path/to/example-group`.

